PR #2, opened 2022-01-31
Replace use of click with Argparse and allow yaml specification of the application

danholdaway:
Replace click with Argparse.

Add new API that we discussed allowing loop over applications and single yaml.

Updated the README with some actual details.

### DIFF
--- a/src/eva/base.py
+++ b/src/eva/base.py
@@ -8,8 +8,10 @@
 
 # imports
 from abc import ABC, abstractmethod
-import click
+import argparse
 import importlib
+import os
+import sys
 import yaml
 
 # local imports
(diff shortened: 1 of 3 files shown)


src/eva/base.py
```diff
@@ -109,19 +111,63 @@ def create_and_run(eva_class_name, config, logger=None):
...
+    parser.add_argument('args', nargs='+', type=str, help='Application name [optional] followed ' +
+                        'by the configuration file [madatory]. E.g. eva ObsCorrelationScatter ' +
+                        'conf.yaml')
+
+
+    args = parser.parse_args()
+    args_list = args.args
+
+    # Make sure only 1 or 2 arguments are present
+    assert len(args_list) <= 2, "The maximum number of arguments is two."
```
CoryMartin-NOAA, line 151: What about if `len(0)`?

CoryMartin-NOAA (commented):
Just one comment/question, otherwise looks good.

src/eva/base.py
danholdaway: I think `argparse` should take care of that with `nargs='+'`, which ensures at least something is provided in the arguments.

CoryMartin-NOAA (approved):
Looks great!

PR #5, opened 2022-02-18
Add in declarative plotting

CoryMartin-NOAA:
## Description

This is probably going to need cleaned up a bit but wanted the bulk of code added early for everyone to look at.


danholdaway:
How straightforward to eliminate all uses of `print` with instead `logger.info`? Or do you not see much purpose for a logger class?

CoryMartin-NOAA:
@danholdaway I think we can do that, sure, that was the intention of this PR, to just mostly copy/paste stuff and then address these sorts of issues

CoryMartin-NOAA:
@danholdaway on closer look all but 1 print in this PR is stuff that we can just not include for now, it was something from Rahul.

danholdaway (approved):

PR #9, opened 2022-02-18
Feature/new api

danholdaway:
## Description

Move to use a single argument API. And clean up the base class quite a bit. 

PR means Yaml changes, sorry! But I think applications as a key word in the Yaml wasn't a good choice and that diagnostics is better and more consistent with the way the code is discuessed.


README.md
```diff
@@ -14,30 +23,28 @@ The eva package can be installed using pip:
...
+diagnostics:
+  - diagnostic name: ObsCorrelationScatter
     ...
-  - application name: ObsMapScatter
+  - diagnostic name: ObsMapScatter
     ...
 ```
 
-eva can also be invoked from another Python module that passes a dictionary, rather than a Yaml file that is later translated into a dictionary. This is achieved as follows:
+eva can also be invoked from another Python module that passes it a dictionary, rather than a Yaml file that is later translated into a dictionary. This is achieved as follows:
```
CoryMartin-NOAA: be consistent here, `yaml` and `Yaml` both in this section, which is best? Or is it `YAML`?

src/eva/eva_base.py
```diff
@@ -0,0 +1,182 @@
...
+        eva_dict = eva_config
+    else:
+        # Create dictionary from the input file
+        with open(eva_config, 'r') as eva_config_opened:
+            eva_dict = yaml.safe_load(eva_config_opened)
+
+    # Get the list of applications
+    try:
+        diagnostic_configs = eva_dict['diagnostics']
+    except Exception:
```
CoryMartin-NOAA: KeyError?

CoryMartin-NOAA (commented):
couple of comments

src/eva/eva_base.py
```diff
@@ -133,7 +133,7 @@ def eva(eva_config, eva_logger=None):
     # Get the list of applications
     try:
         diagnostic_configs = eva_dict['diagnostics']
-    except Exception:
+    except Exception KeyError:
```
CoryMartin-NOAA: Does this run? I was suggesting `except KeyError:`

src/eva/utilities/logger.py
```diff
@@ -42,8 +43,8 @@ def __init__(self, task_name):
 
     def send_message(self, level, message):
 
-        if self.loggerdict[level]:
-            print(level+" "+self.task_name+": "+message)
+        if level == 'ABORT' or self.loggerdict[level]:
```
CoryMartin-NOAA: Can we do a level.upper() here?

src/eva/eva_base.py
danholdaway: No, I made a typo. Will remove

src/eva/utilities/logger.py
danholdaway: Can do

CoryMartin-NOAA:
```suggestion
        if level.upper() == 'ABORT' or self.loggerdict[level]:
```

CoryMartin-NOAA (approved):

setup.py
```diff
@@ -36,6 +36,8 @@
         'pycodestyle>=2.8.0',
         'netCDF4>=1.5.7',
         'matplotlib>=3.4.3',
+        'cartopy>=0.20.2',
```
CoryMartin-NOAA: I think this needs to be downgraded, the current version of cartopy doesn't work with the ubuntu geos library

PR #10, opened 2022-02-18
Fix to install Cartopy dependencies in CI

CoryMartin-NOAA:
## Description

Cartopy has dependencies that are annoying to install. Normally, conda handles this, but ubuntu on GitHub actions allows us to just do `sudo apt-get`


danholdaway (approved):
Ouch, glad I didn't have to figure that one out! Thanks.

PR #11, opened 2022-02-22
Fix colorbar issue in `figure.py`

kevindougherty-noaa:
## Description

This PR fixes the current issue that `figure.py` has with only being able to add a colorbar on every ax. This new feature now can plot a singular colorbar for any type of subplot combination horizontally and vertically. It also gives the user the option to change the location of the colorbar if needed. 

The single colorbar option assumes the user is plotting multiple subplots of the same color series to compare. A note was made in the code that is important to make all objects vmin and vmax the same so there are no comparison errors.

**Potential issue**: this will not be able to plot a single colorbar for specific subplots such as this example [here](https://matplotlib.org/stable/gallery/subplots_axes_and_figures/colorbar_placement.html):
image

Some thought would need to go into restructuring the code if this is something that would need to be used.

### DIFF
--- a/src/eva/plot_tools/figure.py
+++ b/src/eva/plot_tools/figure.py
@@ -145,8 +145,8 @@ class CreateFigure:
             cs = ax.scatter(plotobj.longitude, plotobj.latitude,
                             c=plotobj.data, s=plotobj.markersize,
                             **inputs, transform=self.projection.projection)
-#         if plotobj.colorbar:
-#             self.cs = cs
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):

danholdaway (approved):

slawrence87544 (commented):
These changes seem fine to me, but I have to admit I don't really have a ton of experience with python's colorbar features.

PR #12, opened 2022-02-22
Feature/add unittests

slawrence87544:
## Description

This unit test relies on a new method for loading yaml files
which allows the user to create a substution for an environment
variable identified by the yaml tag !ENVVAR .  If the tag !ENVVAR exists
before a yaml key value, and the environment variable specified between
the {} brackets is set, the yaml load routine will substitude whatever
is set in the specified environment variable.  For instance, if one
set an environment variable 'EVA_TESTS_DIR' to the following:
export EVA_TESTS_DIR=/absolute/path/to/eva/install/location
the the yaml load routine will change the following yaml key

  yaml_key: !ENVVAR ${EVA_TESTS_DIR}/data/datafile.nc4

to

  yaml_key: /absolute/path/to/eva/install/location/data/datafile.nc4


danholdaway:
@slawrence87544 thanks for adding this. Can you update to develop-codespring and use the eva_base.py rather than base.py file? During the code sprint I removed this file and updating things within it.

slawrence87544:
Hi Dan,
I updated my branch with your latest changes which use eva_base.py instead
of base.py.  I re-ran the unit test and it still works.  I noticed that the
config file requirement for the top level key moved back to 'diagnostics'.
Was this intentional?  I'm assuming it is, but I just wanted to confirm
that it was.
-Steve

danholdaway:
Thanks @slawrence87544. I'm a little concerned about the modifications to the base class, which will affect every instance of eva. With the changes we made during the sprint you can now call the `eva_base.eva` method with a dictionary rather than a path to a yaml file. Can you move the methods you added to the base class and put them in `test_obs_correlation_scatter.py` instead? You can convert OBS_CORRELATION_SCATTER_YAML into a dictionary in that routine and then call eva with the dictionary.

danholdaway (approved):
Thank you @slawrence87544 looks good.

PR #17, opened 2022-03-07
Start to add very generic plotting controlled through YAML

CoryMartin-NOAA:
## Description

This is a start to controlling figure generation 100% through YAML. Would like some initial comments on design. 

The tests/config/testDummy.yaml creates a figure like so:
Screen Shot 2022-03-07 at 2 23 00 PM

### DIFF
--- a/src/eva/data/__init__.py
+++ b/src/eva/data/__init__.py
@@ -1,0 +1,9 @@
+# (C) Copyright 2021-2022 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+import os
+
+repo_directory = os.path.dirname(__file__)
--- a/src/eva/data/dummy.py
+++ b/src/eva/data/dummy.py
@@ -6,5 +6,6 @@ class Dummy(EvaBase):
     def execute(self):
         data_dict = {'x': [1,2,3,4,5], 'y': [1,2,3,4,5]}
         self.data = pd.DataFrame.from_dict(data_dict)
+        self.datavars = ['x', 'y']
 
 
--- a/src/eva/data/dummy.py
+++ b/src/eva/data/dummy.py
@@ -1,0 +1,10 @@
+from eva.eva_base import EvaBase
+import pandas as pd
+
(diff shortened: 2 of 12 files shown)


kevindougherty-noaa (commented):
My first initial go through, I like the concept of everything being in a yaml. It's straight forward and you do not need Python experience to utilize this. It will be easy to translate into subplots if necessary and has everything in one spot neatly. However, there will definitely need to be lots of documentation for users on how to include other inputs for features i.e fontsize/fontweight for titles and labels etc.

One thing I'm not quite sure I understand is the data aspect. Will the user be responsible for getting the arrays in order to plot? Do they have to create their own python script? I see `dummy.py` but I'm not sure I entirely understand how that will translate with real data.

CoryMartin-NOAA:
@kevindougherty-noaa the idea is that the first part of the YAML would specify what data would be collected and stored in a dataframe to then be plotted by the second part of the yaml, see https://github.com/danholdaway/eva/blob/feature/plot_schema/src/eva/tests/config/ObsCorrelationScatterDriver.yaml for Dan and I's brainstorming doc (the first part)

kevindougherty-noaa:
@CoryMartin-NOAA I see, that makes sense. There would be a script within this repository that handles grabbing that input and returning the appropriate data?

CoryMartin-NOAA:
yeah the dummy.py Dummy class handles that for now, but a more comprehensive class might be IodaObsSpace or GsiDiag or FV3ModelState for example.

danholdaway:
Looks good to me @CoryMartin-NOAA. I like the design of splitting the creator in base into the data then graphics. It makes a lot of sense. I agree @kevindougherty-noaa that we will need careful documentation. Perhaps we should have a bit of a think about how to document the default Yamls. Probably we need to put lots of info in the docs but making sure we have a system in place to ensure docs updated might be tricky. Once we have this new structure finalized we should do some kind of documentation sprint

CoryMartin-NOAA:
@danholdaway yeah one approach is to have comments for each entry in the default YAML, but is that the best approach? For example, default might not be 'all options'. So it would be good to have a place where all options are listed (and what the defaults are)

CoryMartin-NOAA:
@danholdaway marked ready to review to PR into the develop-reorg branch so that any further changes I make can be minor (and allow you to more easily add the data I/O stuff)

danholdaway (approved):

PR #19, opened 2022-03-21
Add DataContainer class

danholdaway:
## Description

Introduce simple DataCollections class that contains collections of xarray Datasets, accessible by keys. Converts Dummy class to use the new DataCollections class and updates Figure/Scatter example to be YAML driven.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,6 +43,7 @@ setuptools.setup(
         '': [
                'tests/config/*',
                'tests/data/*',
+               'defaults/*.yaml'
              ],
     },
     entry_points={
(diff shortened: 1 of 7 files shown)


CoryMartin-NOAA (approved):
LGTM. Except, am I understanding correctly that the `Dummy` won't work because it is just using `x` and `y`, shouldn't it be like `dummy::x`?

danholdaway:
Oh yeah, good spot. I was updating the Yaml in a sandbox and didn't push back to the repo. Should be updated now.

CoryMartin-NOAA (approved):

PR #20, opened 2022-03-23
Ioda Data Class

danholdaway:
## Description

First PR adding functionality to read IODA files into the collections. Can read conventional and metadata. Not yet tested with radiance data. And need to look at how to remove channels that aren't needed.

I haven't tried making any plots with the data that's read in but this allows some review/discussion of the approach.

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -1,0 +1,97 @@
+# (C) Copyright 2021-2022 NOAA/NWS/EMC
+#
+# (C) Copyright 2021-2022 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+# --------------------------------------------------------------------------------------------------
+
+import xarray as xr
+
+from eva.eva_base import EvaBase
+from eva.utilities.config import get
+
+# --------------------------------------------------------------------------------------------------
+
+
+class IodaObsSpace(EvaBase):
+
+    def execute(self, data_collections):
+
+        # Local copies
+        logger = self.logger
+        config = self.config
+
+        # Get the list of datasets to create
+        datasets = config.get('datasets')
+
+        # Loop over the datasets
+        for dataset in datasets:
+
+            # Get the groups to be read
+            groups = get(dataset, logger, 'groups')
+
+            # Assert the correct use of variables and channels
+            if ('variables' in dataset and 'channels' in dataset):
+                self.logger.abort('Dataset configuraiton for \'' + dataset['name'] + '\'' +
+                                  'contains both channels or variables, should be one or ' +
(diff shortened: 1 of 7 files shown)


src/eva/data/ioda_obs_space.py
```diff
@@ -0,0 +1,96 @@
...
+            if ('variables' in dataset and 'channels' in dataset):
+                self.logger.abort('Dataset configuraiton for \'' + dataset['name'] + '\'' +
+                                  'contains both channels or variables, should be one or ' +
+                                  'the other. If channels present, variable is ' +
+                                  'brightness_temperature')
+
+            # Get channels or variables
+            if ('variables' in dataset):
+                variables = dataset['variables']
+            elif ('channels' in dataset):
```
CoryMartin-NOAA: This is probably okay for now, but there will eventually be cases where 'channels' could also refer to AOD, or antenna temperature, etc. and not just brightness_temperature.

src/eva/data/ioda_obs_space.py
```diff
@@ -0,0 +1,96 @@
...
+                        # likely some problem has occurred
+                        if not ds.keys():
+                            self.logger.abort('Collection \'' + dataset['name'] + ', group ' +
+                                              group + '\' in file ' + filename +
+                                              ' does not have any variables.')
+
+                        # Add the dataset to the collections
+                        data_collections.create_or_add_to_collection(collection_name, ds, 'nlocs')
+
+            print(data_collections)
```
CoryMartin-NOAA, line 84: Is this for debugging or intended to always print?

src/eva/data/ioda_obs_space.py
```diff
@@ -0,0 +1,96 @@
...
+
+            exit()
```
CoryMartin-NOAA, line 86: This is (hopefully) only for debugging :-)

CoryMartin-NOAA (commented):
Looks good to me overall, just a few comments and a few leftover debug lines here before a merge.

src/eva/data/ioda_obs_space.py
danholdaway: Ah yes good point. I'll set it to allow either variables or channels and variables. Is it safe to set it such that if channels are present the variables list must be length 1?

src/eva/data/ioda_obs_space.py
danholdaway: I think this would be a useful thing to print because it shows the users exactly what variables and collections are available. This would help with debugging the next step if the incorrect collection name or variable is chosen.

src/eva/data/ioda_obs_space.py
danholdaway: I'll remove this later on when it's been tested with the plotting software.

src/eva/data/ioda_obs_space.py
CoryMartin-NOAA: I think so, as this would be limited to one obs space. Seems unlikely there would be both something with and without channels in it.

CoryMartin-NOAA (approved):

PR #22, opened 2022-03-29
Add the channel subsetting functionality

danholdaway:
## Description

Remove channels based on user selection

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -12,6 +12,7 @@ import xarray as xr
 
 from eva.eva_base import EvaBase
 from eva.utilities.config import get
+from eva.utilities.utils import parse_channel_list
 
 # --------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


src/eva/data/ioda_obs_space.py
```diff
@@ -64,23 +71,53 @@ def execute(self, data_collections):
                 # Loop over filenames
                 for filename in filenames:
 
-                    # Open file and add collection
-                    with xr.open_dataset(filename, group=group) as ds:
+                    # Read header part of the file to get coordinates
```
CoryMartin-NOAA, line 74: Rahul will say to do a with open like you had before :-)

CoryMartin-NOAA (commented):
One comment about the `with foo as bar` being the 'preferred' way to open files, otherwise looks good to me.

src/eva/data/ioda_obs_space.py
danholdaway: It leads to substantial indenting, which I felt reduced the readability of the code. There are the associated closing of the DataSets later in the code so I don't think it makes too much difference. But I'm happy to return it if you think it's better?

CoryMartin-NOAA: Personally, I think it's fine the way it is. But I know that the 'pythonic' way is the 'with ...'. But leave it as is, readability is important.

danholdaway: OK, will do. We can look again at it. Especially if the code changes as we learn more about the way it works when I start adding filters and trying to make plots.

CoryMartin-NOAA (approved):

PR #24, opened 2022-04-15
Transform infrastructure

danholdaway:
## Description

A few changes to the IodaObsSpace reader class to facilitate the transforms. The main idea of this PR is to demo the intended structure of the transforms. Noting is tested yet. I've added two templates, one for doing arithmetic, i.e. creating omb from the o and b that are read in the data class. And another class for rejecting or accepting certain values based on some criteria. The YAML shows how these classes might work, though the code is not implemented yet.


src/eva/tests/config/testIodaObsSpace.yaml
```diff
@@ -0,0 +1,106 @@
...
+
+
+
+  # Transforms
+  transforms:
+
+    - transform: arithmetic
+      arithmetic type: subtraction
+      collections: [reference]
+      variables: [air_temperature-*, eastward_wind-*]
```
CoryMartin-NOAA, line 57: Is the wildcard here intended to mean 'all' variables (as in shell). Is there a reason why it wouldn't be fine to just do `air_temperature` since the `arithmetic variables` line handles this later?

danholdaway: That might work actually. I think I was trying to imagine situations where the data class has put variables in with certain naming convention. I struggled a bit with how things should be named and it might be something that gets refined as the actual implementation of these transform classes takes shape.

CoryMartin-NOAA: Yeah I think whatever you have that is working is fine for now and we can refine once we have real cases.

CoryMartin-NOAA (approved):

PR #25, opened 2022-05-04
Feature/transforms

danholdaway:
## Description

I think this can be the last PR into the re-org branch before it could be merged into develop. We should be able to reproduce the plotting we have in the GMAO workflow with what would be available with this branch.

- Add the arithmetic transform
- Add an accept where transform
- Update the scatter/figure parts a little bit to accept the new data structures.


danholdaway:
The tests produce the below plots, which show all the locations in black and the locations where the EffectiveQC flag is 0 in red.

aircraft_air_temperature png
amsua_n19-channel5_brightness_temperature png

src/eva/data/data_collections.py
```diff
@@ -66,45 +76,116 @@ def create_or_add_to_collection(self, collection_name, collection, concat_dimens
...
+
+        # This code checks that the naming conventions are compliant with what is expected
+
+        for collection_key in self._collections.keys():
 
-        return self._collections[collection][variable].data
+            # Assert that the collection name does not contain disallowed characters
+            if not string_does_not_contain(disallowed_chars, collection_key):
+                self.logger.abort(f'Collection contains the key \'{collection_key}\', which ' +
+                                  f'contains a characeter that is not permitted ' +
```
CoryMartin-NOAA: typo, `character`

src/eva/data/data_collections.py
```diff
@@ -66,45 +76,116 @@ def create_or_add_to_collection(self, collection_name, collection, concat_dimens
...
 
     # ----------------------------------------------------------------------------------------------
 
-    def __str__(self):
+    def display_collections(self):
 
-        # Print a list of variables that are available in the collection
-        self.logger.info("Collections available: ", )
+        # Display a list of variables that are available in the collection
+        self.logger.info('-'*80)
```
CoryMartin-NOAA, line 183: Does this print 80 `-`? I've never known about this syntax before!

src/eva/diagnostics/scatter.py
```diff
@@ -1,23 +1,55 @@
...
     def __init__(self, config, logger, dataobj):
-        # prepare data based on config
+
+        # Get the data to plot from the data_collection
+        # ---------------------------------------------
         varnames = config['comparison']
-        xdata = dataobj.get_variable_data(varnames[0])
-        ydata = dataobj.get_variable_data(varnames[1])
-        # create declarative plotting Scatter object
+        var0 = varnames[0]
```
CoryMartin-NOAA, line 16: FYI, I have in my branch a change that instead of `comparison` and then a list of vars it is now two keys `x:` and `y:` which I think is more explicit for scatter plots.

src/eva/utilities/utils.py
```diff
@@ -15,6 +17,15 @@
 # --------------------------------------------------------------------------------------------------
 
 
+class fontColors:
```
CoryMartin-NOAA, line 20: very nice!

CoryMartin-NOAA (commented):
One small typo, otherwise this all looks fantastic!

src/eva/data/data_collections.py
danholdaway: It does. The wonderful world of Python.

src/eva/diagnostics/scatter.py
danholdaway: OK, good to know. We can update the YAMLs. That makes more sense.

src/eva/utilities/utils.py
danholdaway: I did wonder if this might be unpredictable say on Windows. But I'm guessing there might be other problems trying to run on Windows.

src/eva/data/data_collections.py
danholdaway: Done

PR #26, opened 2022-05-20
Create nlocs using range

CoryMartin-NOAA:
With recent (all new?) IODA output files, nlocs is all zeros, which causes issues when combining xarray datasets.

This PR will check if the max value of `nlocs` is zero for ioda_obs_space, if so, it will use `range` combined with `nlocs.size` and add 0 ... nlocs to the 0 values so that there are unique values for each location.

Tested on Orion and it fixes my issue. See the plot of AMSU-A using output from a recent run of H(x) from 'GDASApp'.
Screenshot from 2022-05-20 20-36-14

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -40,6 +40,7 @@ def subset_channels(ds, channels, add_channels_variable=False):
 
 # --------------------------------------------------------------------------------------------------
 
+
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
Unrelated to this PR, but this plot needs (nobs) for both with and without QC added to the legend I think

danholdaway (approved):

PR #28, opened 2022-06-02
Upgrades for integration with GMAO workflow

danholdaway:
## Description

Biggest change here is to inroduce some basic batch functionally into the FigureDriver class. But also a number of other small changes that have been needed to integrate with the GMAO workflow.

- Add `nan_float_values_outside_threshold` routine for screening out large values not captured by missing value metadata. Probably could be cleaner and use the actual missing value information but for future work.
- Improve the print method for the data class with min/max/rms to give users a better impression of what the data contains.
- Use `mask_and_scale` in opening datasets to prevent integers being turned to floats. Not sure why xarray does that by default. I guess to screen for missing values.
- Use `decode_times` in opening datasets to improve compatibility with different ways the datetime information might be written.
- Default to reading all variables from the selected groups. Helps with metadata, which differ for different instruments.
- Some more verbosity in the base class.
- Close figures to prevent memory overload when batching.
- Allow for path in `savefig` name.
- Remove deprecated argument passed to `savefig`.
- Make the color of the regression curves match the color of the scatters.
- Remove the dummy code put in for testing.
- Remove old obs correlation scatter code.
- Update testing files
- Improve readability of the transforms YAML.
- Set version ahead of tagging after develop-reorg is merged into develop.

jedi_hofx_vs_obs_gmi_gpm_brightness_temperature5


src/eva/data/data_collections.py
```diff
@@ -177,15 +181,79 @@ def validate_names(self):
...
+                # For float data sceen outside threshold
+                if 'float' in str(data_var_value.dtype):
+                    data_var_value[np.abs(data_var_value) > threshold] = np.nan
+
+    # ----------------------------------------------------------------------------------------------
+
     def display_collections(self):
 
+        minmaxrms_format_dict = {}
+        minmaxrms_format_dict['float32'] = '{:+.4e}'
```
CoryMartin-NOAA: Suggest this syntax/format instead for including additional types in the future:
```
minmaxrms_format_dict = {
    'float32': '{:+.4e}',
}
```

src/eva/data/data_collections.py
```diff
@@ -177,15 +181,79 @@ def validate_names(self):
...
+        minmaxrms_format_dict['int32'] = '{:+11d}'
+
         # Display a list of variables that are available in the collection
-        self.logger.info('-'*80)
+        self.logger.info('-'*100)
```
CoryMartin-NOAA: why 100, 80 is the standard default width of a terminal isn't it?

CoryMartin-NOAA (approved):
Two minor comments but nothing that is a show-stopper. Thanks for all these great additions, especially the batch plotting!

src/eva/data/data_collections.py
danholdaway: Happy to go back to 80. It's just a way of seeing where one object starts and ends as it's printed. By the way the logger will insert some text in front of whatever you set.

src/eva/data/data_collections.py
danholdaway: Thanks, that's more compact.

PR #30, opened 2022-06-07
feature/latlon-dan -> feature/latlon

danholdaway:
## Description

Merge develop and perform a bit of clean up. Map plot looks good for the surface GeoVaLs, thanks for adding this.


CoryMartin-NOAA:
Would it be better to just merge this into develop? It should work well enough for now, right? I'm okay with that if you are, keeps the reviews more "agile". Otherwise this is fine too.

danholdaway:
Sure, I'm fine with that. Didn't want to commandeer your efforts in case you had more work in mind but you can always make a new branch. I changed the base

CoryMartin-NOAA:
haven't most of the changes shown here been merged in already? Is this just a github artifact?

CoryMartin-NOAA (commented):
Disregard my previous comment, I see the relevant changes only now

CoryMartin-NOAA (approved):

PR #31, opened 2022-06-09
Bugfix/zero len data

danholdaway:
## Description

In situations where e.g. no data passes the QC filters the regression will crash. This adds a check that the data has some length before computing the regression.


CoryMartin-NOAA (approved):
Good catch, thanks!

PR #34, opened 2022-06-22
Modify regex split

### LINKED ISSUE #33: How to deal with 'mean' in a variable name
If for example, someone wants:

` experiment::ObsValue::${variable}-experiment::hofx_y_mean_xb0::${variable}`

like from the output from the ensemble application(s), eva will fail.

It finds 'mean' and tries to separate it out, even though it is part of the variable name. Can we 'escape' a keyword in the variable name somehow?

Thanks to @weihuang-jedi for finding this.

CoryMartin-NOAA:
Closes #33 

This removes `mean`, `sqrt`, and `\d` from the `regex.split` command for the arithmetic.

Currently, if a variable/group/collection contains a number or 'mean' or 'sqrt' in its name, it will fail. This should fix that. However, we will still need a solution eventually for dealing with 'functions' (mean, sqrt, others)

### DIFF
--- a/.github/workflows/codestyle.yml
+++ b/.github/workflows/codestyle.yml
@@ -13,6 +13,7 @@ jobs:
       run: |
         sudo apt-get install libproj-dev proj-data proj-bin
         sudo apt-get install libgeos-dev musl-dev libc-dev
+        sudo apt-get install libcurl4-openssl-dev
         sudo ln -s /usr/lib/x86_64-linux-musl/libc.so /lib/libc.musl-x86_64.so.1
         mkdir proj && cd proj
         curl -sSL https://download.osgeo.org/proj/proj-8.2.1.tar.gz | tar -xvz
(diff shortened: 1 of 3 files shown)


CoryMartin-NOAA:
There's some bizarre stuff going on with the CI attempting to install cartopy... which is an old version because of an old version of 'proj'. I think we probably need to get new 'proj' compiled/installed in the CI and use a newer cartopy. I have to go to meetings but can investigate later today.

CoryMartin-NOAA:
@danholdaway since we don't have unit tests, should the CI that checks coding norms just do that and not install dependencies?

danholdaway:
Yeah that would likely make sense for now. Just turn off all this install. And then we wont have to have this strange dependency on a fixed version.

danholdaway (approved):

CoryMartin-NOAA:
@danholdaway well I found a fix for the short term that keeps things as-is

PR #40, opened 2022-08-05
Dynamic config and additional testing

danholdaway:
## Description

- Add option to define aspects of the config based on the fields to be plotted.
- Update the test YAMLs
- Add a new CI pipeline that runs the plotting software (application tests).
- Add a new CI pipeline to ensure quality of YAMLs
- Clean up the naming of all the CI.

Users can run the tests that are run in GitHub actions with:

1. Clone eva with git-lfs enabled.
2. Install eva somewhere
3. `eva_tests application`


danholdaway:
Darn. CI was all passing just now. I changed some YAMLs and now it is failing. Will take a look

src/eva/plot_tools/dynamic_config.py
```diff
@@ -0,0 +1,132 @@
...
+    varname_cgv = varname.split('::')
+    datavar = data_collections.get_variable_data(varname_cgv[0], varname_cgv[1],
+                                                 varname_cgv[2], channel)
+
+    # Reorder the data array
+    datavar = np.sort(datavar)
+
+    # Decide how many values to throw out on each end of the dataset
+    n = datavar.size
+    n_throw_out = (100-percentage_capture) * n / 200
```
CoryMartin-NOAA: Must be Friday PM brain, but why divide by 200? Is this 'half' and going from 1% to 0.01? If so, suggest changing this to:
`n_throw_out = ((1-(percentage_capture/100)) * n ) / 2`
Or scale it after line 28.

CoryMartin-NOAA (commented):
One minor comment, will do another review/approve once tests pass

src/eva/plot_tools/dynamic_config.py
danholdaway: I changed to line to make the divide by 2 more explicit.

danholdaway:
@CoryMartin-NOAA after much back and forth I have a solution to the CI that seems to work. It involves setting the cartopy version to 0.18.0 and installing everything using conda. All should be well until someone wants a feature from a newer version of cartopy. Hopefully by then apt will have a more modern version of proj. From the work the last coupe of days trying to get this to work it makes me wonder about the general stability of cartopy. Likely a dependency we don't have a choice over but perhaps something to keep an eye on. It has what seem to be a not well supported set of dependencies, that can be complex to install outside of situations where it is managed for us.

CoryMartin-NOAA:
@danholdaway yes this is the downside of cartopy being a MetOffice creation, it doesn't have quite the community support that many other Python libraries (matplotlib, numpy, etc.) have. I think mapping is in general a complicated thing, so this isn't terribly surprising. NOAA is using MET+ for verification, I am 95% sure cartopy is required for MET+, so that means it won't be a DA problem for install. Are there needs on discover/NASA HPC that will require cartopy?

I'll do a final review later this morning.

src/eva/plot_tools/dynamic_config.py
```diff
@@ -0,0 +1,132 @@
...
+
+    # Find minimum and maximum values
+    cmap = option_dict.get('sequential colormap', 'viridis')
+    vmax = np.nanmax(datavar_check)
+    vmin = np.nanmin(datavar_check)
+
+    # If positive and negative values are present then a diverging colormap centered on zero should
+    # be used.
+    if vmin < 0.0 and vmax > 0.0:
+        vmax = np.nanmax(np.abs(datavar_check))
```
CoryMartin-NOAA, line 63: What if the max value is 2 and the min value is -50? We need to compute min and max and then get the max of the abs value of each to use as the colorbar endpoints.

src/eva/plot_tools/dynamic_config.py
```diff
@@ -0,0 +1,132 @@
...
+    varname_cgv = varname.split('::')
+    datavar = data_collections.get_variable_data(varname_cgv[0], varname_cgv[1],
+                                                 varname_cgv[2], channel)
+
+    # Compute size of the array of data
+    n = np.count_nonzero(~np.isnan(datavar))
+
+    # Check for zero data, set to 3 as a reasonable minimum
+    if n == 0:
+        n = 2
```
CoryMartin-NOAA: 2? but comment says 3

src/eva/plot_tools/dynamic_config.py
danholdaway: In that case wouldn't the result of `np.abs(datavar_check)` be 50? Then `vmax` would be 50 and `vmin` -50.

src/eva/plot_tools/dynamic_config.py
danholdaway: Good spot, it should be 3. Will change.

src/eva/plot_tools/dynamic_config.py
CoryMartin-NOAA: ah, yeah, duh.

danholdaway:
> @danholdaway yes this is the downside of cartopy being a MetOffice creation, it doesn't have quite the community support that many other Python libraries (matplotlib, numpy, etc.) have. I think mapping is in general a complicated thing, so this isn't terribly surprising. NOAA is using MET+ for verification, I am 95% sure cartopy is required for MET+, so that means it won't be a DA problem for install. Are there needs on discover/NASA HPC that will require cartopy?
> 
> I'll do a final review later this morning.

I'm sure people are using cartopy in lots of places and I believe Met+ is part of the planning. Interestingly I had no problem at all installing it on Discover or my personal machine. Part of the issue might just be the inflexibility inherent in GitHub actions. Unless I'm missing something it's much harder to play around with things to test out. Each little change is a push and wait. So you run into a problem like using conda to install cartopy and it forces version 0.18.0. Why? On a local machine you could play around for a while to try and figure it out. But in the container on the GitHub servers that becomes hours of work.

CoryMartin-NOAA:
@danholdaway we can use a docker container that already has this installed, if there is one. https://github.com/CoryMartin-NOAA/plotioda/blob/main/.github/workflows/tests.yaml for example, I used a JCSDA docker and build/cache ioda to then use with the ioda python api.

danholdaway:
Corrected. Hopefully it will make the future effort of reviewing the code a bit simpler now we know the plots are being generated in the CI. I wonder if the next level of CI could involve positing the plots to the issue or a website somehow.

CoryMartin-NOAA (approved):

PR #45, opened 2022-09-07
Ensure sensor_channels not local channels are used when subsetting

CoryMartin-NOAA:


CoryMartin-NOAA:
work done by @ADCollard

CoryMartin-NOAA:
@ADCollard fix coding norms!
```
./src/eva/data/ioda_obs_space.py:101:32: E225 missing whitespace around operator
./src/eva/data/ioda_obs_space.py:146:33: E225 missing whitespace around operator
```

kevindougherty-noaa:
Looks good once you fix code norms!

src/eva/data/ioda_obs_space.py
```diff
@@ -97,6 +97,9 @@ def execute(self, data_collections):
                 # Read header part of the file to get coordinates
                 ds_groups = xr.Dataset()
 
+                # Save sensor_channels for later
+                sensor_channels = ds_header['nchans']
```
danholdaway: I think this code might need to be protected to ensure it's not used when `nchans` is not in the file.

danholdaway (changes requested):
One comment, that I think is why there's a failure of the application test.

src/eva/data/ioda_obs_space.py
CoryMartin-NOAA: Yeah so then the first one would need:
```
has_sensor_channels = False
if 'nchans' in ds_header.keys():
    sensor_channels = ds_header['nchans']
    has_sensor_channels = True
```
and then the later line:

```
if has_sensor_channels:
    ds['nchans'] = sensor_channels
```

Or some other similar solution

ADCollard:
@CoryMartin-NOAA, the coding norms are fixed but for some reason the eva application test is failing for aircraft.  So I will investigate further.

src/eva/data/ioda_obs_space.py
danholdaway:
```suggestion
sensor_channels = 0
if 'nchans' in ds_header.keys():
    sensor_channels = ds_header['nchans']
```

danholdaway:
> @CoryMartin-NOAA, the coding norms are fixed but for some reason the eva application test is failing for aircraft. So I will investigate further.

I just put in a suggested change to protect the new code.

src/eva/data/ioda_obs_space.py
danholdaway: And that works too.

src/eva/data/ioda_obs_space.py
```diff
@@ -98,7 +98,9 @@ def execute(self, data_collections):
                 ds_groups = xr.Dataset()
 
                 # Save sensor_channels for later
-                sensor_channels = ds_header['nchans']
+sensor_channels = 0
```
CoryMartin-NOAA: @ADCollard great indentation! :-)

ADCollard: @CoryMartin-NOAA Not sure what happened there, I accepted some suggested changes and it ended up like that.

src/eva/data/ioda_obs_space.py
```diff
@@ -139,6 +144,10 @@ def execute(self, data_collections):
                         rename_dict[group_var] = group_name + '::' + group_var
                     ds = ds.rename(rename_dict)
 
+                    # Reset channel numbers from header
+                    if 'nchans' in ds.keys()
```
CoryMartin-NOAA: I'm not sure this will work, have you tested it? Because `ds.keys()` is for the group which may not contain the header information? But I'm not sure.

danholdaway (approved):

Issue #47, opened 2022-09-09
Add statistics for each channel


ADCollard:
It would be useful to be able to produce statistics for each channel, especially for instruments like IASI with thousands of channels.

I have added a new routine to [my fork](https://github.com/ADCollard/eva) called [stats2d.py](https://github.com/ADCollard/eva/blob/develop/src/eva/transforms/stats2d.py) that collects statistics along the` nobs `direction only, producing channel by channel stats for radiance measurements.

The code will automatically append the default statistic names to the variable.  E.g.:
```
INFO DataCollections:   hofx::brightness_temperature                           (float32) | Min=+1.8602e+02, Max=+2.9961e+02, RMS=+2.4154e+02
hofxMean::brightness_temperature                       (float32) | Min=+2.1379e+02, Max=+2.8392e+02, RMS=+2.4133e+02
INFO DataCollections:   hofxStd::brightness_temperature                        (float32) | Min=+5.5126e+00, Max=+1.6265e+01, RMS=+1.0095e+01
INFO DataCollections:   hofxCount::brightness_temperature                      (int64  ) | Min=+5.5126e+00, Max=+1.6265e+01, RMS=+1.0095e+01
INFO DataCollections:   hofxMedian::brightness_temperature                     (float32) | Min=+2.1240e+02, Max=+2.8923e+02, RMS=+2.4330e+02
INFO DataCollections:   hofxMin::brightness_temperature                        (float32) | Min=+1.8602e+02, Max=+2.2767e+02, RMS=+2.0081e+02
INFO DataCollections:   hofxMax::brightness_temperature                        (float32) | Min=+2.3110e+02, Max=+2.9961e+02, RMS=+2.5440e+02
```
The YAML to do this is simple:
```
  transforms:

    # Stats for hofx
    - transform: stats2d
      variable_name: experiment::hofx::${variable}
      for:
        variable: *variables
```

A pull request will be issued soon.

ADCollard:
A sample plot using this capability:
![Screen Shot 2022-09-09 at 17 44 28 PM](https://user-images.githubusercontent.com/40322596/189450069-e5916c27-004f-46e4-9b52-727b2e127413.png)

ADCollard:
Changes have been merged into deveop

PR #48, opened 2022-09-12
Channel stats

ADCollard:
## Description

This PR adds the ability to produce statistics on a channel-by-channel basis.  See #47 for more information.

## Dependencies

None

## Impact

No impact on other repositories


src/eva/tests/config/testIodaObsSpaceIASI_Metop-A.yaml
```diff
@@ -0,0 +1,90 @@
...
+# on Orion-login-4.HPC.MsState.Edu at 2022-08-31T10:58:39Z
+# template YAML file to create EVA YAML files
+# based on obs spaces listed in JEDI YAML files
+diagnostics:
+- data:
+    type: IodaObsSpace
+    datasets:
+      - name: experiment
+        filenames:
+          - ${data_input_path}/iasi_metop-a.hofx.2021-08-01T00:00:00Z.nc4
```
CoryMartin-NOAA, line 12: This file is 1.6MB, how many locations is it? Can we subset it some for the test?

src/eva/transforms/channel_stats.py
```diff
@@ -0,0 +1,62 @@
...
+# --------------------------------------------------------------------------------------------------
+
+
+def channel_stats(config, data_collections):
+
+    # Create a logger
+    logger = Logger('Stats2DTransform')
+
+    # Define stat functions to calcualte
+    stat_functions = ['Mean', 'Std', 'Count', 'Median', 'Min', 'Max']
```
CoryMartin-NOAA: Do we ever want to specify the list of stats or always compute these 6?

CoryMartin-NOAA (commented):
2 small comments/questions

CoryMartin-NOAA:
@ADCollard both the YAML norms and the Python norms tests failed. Do you know how to view the output on Github?

ADCollard:
> @ADCollard both the YAML norms and the Python norms tests failed. Do you know how to view the output on Github?

@CoryMartin-NOAA Yes, fixing them now.

src/eva/tests/config/testIodaObsSpaceIASI_Metop-A.yaml
CoryMartin-NOAA: Ok I see it is 101 locations, that's probably as good as we can subset it

src/eva/transforms/channel_stats.py
ADCollard: @CoryMartin-NOAA I thought this was the most user-friendly way of going about it.   But are you suggesting I add the capability to specify stat_functions in the YAML with this as a default?

CoryMartin-NOAA: That is my suggestion but not a requirement if you don't think it is necessary.

src/eva/tests/config/testIodaObsSpaceIASI_Metop-A.yaml
ADCollard: @CoryMartin-NOAA To be honest, even 1 or 2 obs would be fine for this purpose.  How about I subset to 10 obs so that the file size is similar to AMSU-A?

danholdaway (approved):
Thanks. This is another great feature to have. In the plots in the issue there is still a linear regression. For this kind of plot can it be turned off @CoryMartin-NOAA? I never know how to turn it off. Maybe it's for another PR.

CoryMartin-NOAA:
@danholdaway yeah @ADCollard and I already discussed it. It's that the default includes it, and because the way it is called, it cannot be overwritten. So we need to modify the code that checks for linear regression in YAML and instead check if it is `true` not that if the key exists.

danholdaway:
> @danholdaway yeah @ADCollard and I already discussed it. It's that the default includes it, and because the way it is called, it cannot be overwritten. So we need to modify the code that checks for linear regression in YAML and instead check if it is `true` not that if the key exists.

OK. Do you want to make an issue so it's not forgotten and then we can merge this PR (once you resolve the above conversation) and tackle in a future PR?

CoryMartin-NOAA:
https://github.com/danholdaway/eva/issues/49 

@ADCollard are you further subsetting the test file? If not, we can merge.

ADCollard:
> #49
> 
> @ADCollard are you further subsetting the test file? If not, we can merge.

@CoryMartin-NOAA I have subset the file, but have not pushed it as I was working on adding the option to specify which stats are calculated. Or should I do that in a separate PR?

CoryMartin-NOAA:
@ADCollard oh duh, forgot about that part already. Adding it to this PR is good. I will 'hold my horses' until that is made optional.

ADCollard:
> @ADCollard oh duh, forgot about that part already. Adding it to this PR is good. I will 'hold my horses' until that is made optional.

@CoryMartin-NOAA I have committed the small IASI data file and it is not possible to specify a subset of statistical calculations (but it defaults mean, median, std, min and max).

src/eva/transforms/channel_stats.py
```diff
@@ -0,0 +1,67 @@
...
+
+def channel_stats(config, data_collections):
+
+    # Create a logger
+    logger = Logger('ChannelStatsTransform')
+
+    # Define default stat functions to calculate
+    if 'statistic list' in config:
+        stat_functions = get(config, logger, 'statistic list')
+        print('stat_functions=',stat_functions)
```
CoryMartin-NOAA: Use the logger instead?

ADCollard: @CoryMartin-NOAA That print statement was left from debugging.   I have removed it.

src/eva/transforms/channel_stats.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def channel_stats(config, data_collections):
+
+    # Create a logger
+    logger = Logger('ChannelStatsTransform')
+
+    # Define default stat functions to calculate
+    if 'statistic list' in config:
+        stat_functions = get(config, logger, 'statistic list')
```
CoryMartin-NOAA, line 35: Last comment and then I'll leave @ADCollard alone. Do we want to check if the input list is 'valid'? Either here by verifying that the list contains valid values. Or I suggest on line ~59, should have a try/except or some sort of check so it doesn't fail completely? (Or do we want it to fail completely?)

ADCollard: @CoryMartin-NOAA I think it should fail completely.  You want to be sure your YAML is right and does what you want rather than just ignoring all your mistakes.

I guess I can add some code to make it fail more elegantly, but inscrutable messages when you get the YAML wrong seems to be the norm as far as I can tell.  :-)

CoryMartin-NOAA: Roger that.

PR #50, opened 2022-09-16
Adding missing value threshold option

asewnath:
## Description

Adding in the missing value threshold option in the yaml files so that users can specify the field. If the optional field is missing, a default value is used.

Provide a detailed description of what this PR does.

Made changes in ioda_obs_space.py to create optional missing value threshold field. Adding tornado module to setup.py since it is required for certain functionality and is not already loaded on some machines.


danholdaway:
Thanks @asewnath. Can you take out the change to setup.py adding tornado? I know you need this to compile locally but unfortunately I think you're the only one and we don't want to insist on everyone needing it.

danholdaway (approved):

PR #51, opened 2022-09-22
Timing

danholdaway:
## Description

Add a basic timer that can be used to profile parts of the code. At the end of the run the statistics will be printed in the following format:

Screen Shot 2022-09-22 at 11 25 51 AM

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -123,7 +123,7 @@ class IodaObsSpace(EvaBase):
                     # Read the group
                     timing.start(f'IodaObsSpace: open_dataset')
                     ds = open_dataset(filename, group=group_name, mask_and_scale=False,
-                                         decode_times=False)
+                                      decode_times=False)
                     timing.stop(f'IodaObsSpace: open_dataset')
 
                     # If user specifies all variables set to group list
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -125,10 +125,10 @@ class IodaObsSpace(EvaBase):
                     collection_name = dataset['name']
 
                     # Read the group
-                    timing.start(f'IodaObsSpace: open_dataset')
+                    timing.start(f'IodaObsSpace: open_dataset {os.path.basename(filename)}')
                     ds = open_dataset(filename, group=group_name, mask_and_scale=False,
                                       decode_times=False)
-                    timing.stop(f'IodaObsSpace: open_dataset')
+                    timing.stop(f'IodaObsSpace: open_dataset {os.path.basename(filename)}')
 
                     # If user specifies all variables set to group list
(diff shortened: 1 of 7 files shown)


src/eva/data/ioda_obs_space.py
```diff
@@ -125,8 +125,10 @@ def execute(self, data_collections):
                     collection_name = dataset['name']
 
                     # Read the group
-                    ds = xr.open_dataset(filename, group=group_name, mask_and_scale=False,
-                                         decode_times=False)
+                    timing.start(f'IodaObsSpace: open_dataset')
```
CoryMartin-NOAA: Thoughts on adding the filename to the timer (perhaps a debug option?) in case there are multiple files to see which one is slow to read?

src/eva/eva_base.py
```diff
@@ -91,32 +92,38 @@ def create_eva_object(self, eva_class_name, eva_group_name, config, eva_logger):
 
         # Import class based on user selected task
         # ----------------------------------------
+        module_to_import = "eva."+eva_group_name+"."+eva_module_name
+        # timing.start(f'EvaFactory import: {eva_class_name} from {module_to_import}')
```
CoryMartin-NOAA: Why are these commented out?

src/eva/utilities/timing.py
```diff
@@ -0,0 +1,141 @@
...
+                              f'initialized.')
+
+        # Check that the timer is in fact running
+        if not self.timing_dict[timer_name]['running']:
+            self.logger.abort(f'Trying to stop the timer for \'{timer_name}\' but it is not ' +
+                              f'running')
+
+        # Record the final time
+        elapsed = time.perf_counter() - self.timing_dict[timer_name]['start_time']
+        self.timing_dict[timer_name]['total_time'] = elapsed
```
CoryMartin-NOAA: what if you start, stop, then start again? Did I miss where the elapsed time is added to the starting total time?

CoryMartin-NOAA (commented):
few questions/comments but mostly looks great!

src/eva/data/ioda_obs_space.py
danholdaway: I've added the filename. It needed a bit of work to allow for longer strings without ruining the look of the output but I think it's OK. Here's an example of what it looks like now:

Screen Shot 2022-09-23 at 1 54 57 PM

src/eva/eva_base.py
danholdaway: With the longer filenames allowed (referenced above) I've uncommented these.

src/eva/utilities/timing.py
danholdaway: Good catch, thanks. I think it's fixed now.

CoryMartin-NOAA (approved):
looks great, thanks for adding this!

PR #52, opened 2022-09-22
TryToFixCi

danholdaway:
## Description

Try to fix the CI failures

### DIFF
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -1,7 +1,7 @@
-pyyaml>=6.0
-pycodestyle>=2.8.0
-netCDF4>=1.5.3
+pyyaml==6.0
+pycodestyle==2.8.0
+netCDF4==1.5.3
 matplotlib==3.5.2
 cartopy==0.18.0
-scikit-learn>=1.0.2
-xarray>=0.11.3
+scikit-learn==1.0.2
+xarray==0.11.3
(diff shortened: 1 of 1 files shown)


danholdaway:
OK that was easier than I expected it to be. @kevindougherty-noaa this might be what you need to do in EMCpy too if you didn't get to a workaround yet. All comes back to the fact that we can't install a recent version of Cartopy because the GitHub servers use an old proj. Now we have to fix to an old version of matplotlib too so it's compatible with the old version of cartopy.

kevindougherty-noaa (approved):
Excellent! I'll give it a try on EMCPy as well

danholdaway:
I went ahead and fixed the versions of all packages so we don't experience another break to CI. I'm not sure if this is a good thing to do or not since it likely means the CI environment will become quite different from the environment people will work with in practice. At some point someone will want to use a new feature that isn't available in the CI and then we'll have to revisit all this. I guess some testing is better than no testing. Maybe in the long term we should aim to use the spack-stack container in the action.

CoryMartin-NOAA:
I've used a JCSDA docker container in github actions before for IODA and it was quite straightforward. If they have a spack-stack container available, we should move to that soon.

danholdaway:
Dom is working on a spack-stack container but I don't think it's ready yet.

PR #53, opened 2022-09-23
Add density plot and make doing the regression flexible

danholdaway:
## Description

1. Add density plot using `seaborn`. Keeping `seaborn` as an optional dependency.
2. Add option to turn off linear regression plot for scatters.
3. Add zero line to histograms

*Note that these changes would need merging to EMCpy*

<ins>New IASI mean h(x) vs Channel plot produced by the tests without regression line:</ins>

meanhofx_vs_channel

<ins>New density plot produced by the tests:</ins>

gsi_omb_jedi_omb_density_amsua_n19_brightness_temperature_8

<ins>Zero line on histograms:</ins>

gsi_omb_jedi_omb_histogram_amsua_n19_brightness_temperature_8


setup.py
```diff
@@ -36,7 +36,7 @@
         'pycodestyle>=2.8.0',
         'netCDF4>=1.5.3',
         'matplotlib>=3.5.2',
-        'cartopy>=0.20.2',
+        'cartopy>=0.18.0',
```
CoryMartin-NOAA, line 39: ugh we have to revert again? Doesn't this reversion fix all of the proj issues?

src/eva/defaults/scatter.yaml
```diff
@@ -8,6 +8,7 @@ alpha:
 linewidths: 1.5
 edgecolors:
 label: 'Points'
+do_linear_regression: True
```
CoryMartin-NOAA, line 11: thank you for fixing this

CoryMartin-NOAA (approved):
Looks good, I'll work with @kevindougherty-noaa to get this in emcpy soon. We should try to freeze and get this sorted out ASAP so we don't need to keep duplicating effort.

setup.py
danholdaway: I just figured I would set the minimum to this version to eliminate the issues for people who have that old version of proj. But I think we should still have the requirements GitHub file that fixes the versions. That way as packages get updated we don't see the breakage where new versions of packages don't realize that cartopy is tethered to some old thing. That was what caused the break to CI in Eva and EMCpy

Issue #57, opened 2022-10-20
Add `emcpy` as submodule to replace current plotting scripts


kevindougherty-noaa:
## Description

All of the plotting code has been moved over to `EMCPy` to be managed in it's own repository. Rather than have efforts duplicated in `eva`, it is best to import `EMCPy` as a submodule and utilize the plotting codes from there. 

Adding `EMCPy` as a submodule is not necessarily a difficult task, but pointing to the right paths to allow the imports to work properly is proving to be a bit difficult.

PR #58, opened 2022-10-20
Add EMCPy as a submodule

kevindougherty-noaa:
## Description

PR related to Issue #57. Currently, I have added EMCPy as a submodule and updated `setup.py` so that eva will build.

`eva` imports still work, however, when trying to point to any `emcpy` modules, some issues arise. The first being issues with `__init__.py` files from `emcpy`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setuptools.setup(
         'cartopy>=0.18.0',
         'scikit-learn>=1.0.2',
         'xarray>=0.11.3',
-        'emcpy @ git+https://github.com/NOAA-EMC/' + 
+        'emcpy @ git+https://github.com/NOAA-EMC/' +
         'emcpy@8c1958e9ba8906a35a9677030ffa494519e11489#egg=emcpy',
     ],
     package_data={
--- a/setup.py
+++ b/setup.py
@@ -16,13 +16,12 @@ import setuptools
 
 setuptools.setup(
     name='eva',
-    version='1.2.2',
+    version='1.3',
     author='Community owned code',
     description='Evaluation and Verification of an Analysis',
-    url='https://github.com/danholdaway/eva',
-    package_dir={'emcpy': './emcpy/src/emcpy',
-                 '': 'src'},
-    packages=setuptools.find_packages(),
+    url='https://github.com/JCSDA-internal/eva',
+    package_dir={'': 'src'},
+    packages=setuptools.find_packages('src'),
     classifiers=[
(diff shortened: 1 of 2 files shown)


danholdaway:
I would have expected to see all the EMCpy code being removed. Is the plan for a two-step process here?

kevindougherty-noaa:
I have figured out how to add EMCPy to download with EVA. I tried to use git submodules, however it proved to be really difficult. Adding it as an install requirement pointing at a specific git commit sets everything up so we can import emcpy modules instead of eva for plotting.

I am not sure why the eva applications tests are failing.

CoryMartin-NOAA:
@danholdaway yeah, I told @kevindougherty-noaa it would be easier to review if we do this in 2 steps so we can ensure the dependency is installed properly first, and then change the code. Is that okay with you?

The test failures is interesting. Perhaps the install isn't working properly?

kevindougherty-noaa:
@danholdaway Cory suggested we have this PR just change the `setup.py` and then follow with a PR to change all the code.

danholdaway:
Thanks for working on this @kevindougherty-noaa. This solution actually seems more elegant to me than Git submodules. Just let Pip handle things for you! I'll try to fix the CI. It doesn't have anything to do with your PR

setup.py
```diff
@@ -39,6 +39,8 @@
         'cartopy>=0.18.0',
         'scikit-learn>=1.0.2',
         'xarray>=0.11.3',
+        'emcpy @ git+https://github.com/NOAA-EMC/' +
```
danholdaway, line 42: Is this cloning a specific hash? I wonder if we actually want develop?

CoryMartin-NOAA: I feel like using develop can be dangerous, a specific hash ensures things don't break. But we can certainly switch to develop if you have a preference for that.

danholdaway: Yeah it's a good point and one of those situations without a satisfactory solution. If we use a hash we might find that at some point eva and emypy develop are no longer working together, potentially with a lot of things needing changing in Eva. If we use develop we might be forever breaking the CI. I'll defer to you guys on what you think is best. For sure we would want tags of Eva to use hashes of emcpy.

danholdaway (approved):

CoryMartin-NOAA (approved):

setup.py
CoryMartin-NOAA: Yeah I think as long as we don't let eva's hash of emcpy get too stale this should be the way to go. It ensures things do not break for users.

PR #60, opened 2022-10-27
Import emcpy modules

kevindougherty-noaa:
## Description

Replaced all plotting dependency codes within `eva` to be imported by `emcpy`.

This will close issue #57


src/eva/diagnostics/density.py
```diff
@@ -1,7 +1,7 @@
 from eva.eva_path import return_eva_path
 from eva.utilities.config import get
 from eva.utilities.utils import get_schema, update_object, slice_var_from_str
-import eva.plot_tools.plots
+from emcpy.plots.plots import Density as DensityPlot
```
CoryMartin-NOAA: Is there a reason why not just keeping it as `emcpy.plots.plots.Density`? Or is that too verbose?

kevindougherty-noaa: I tried that, and it fails. I get the following error:
`ModuleNotFoundError: No module named 'emcpy.plots.plots.Scatter'; 'emcpy.plots.plots' is not a package`

CoryMartin-NOAA: No wouldn't it be `import emcpy.plots.plots`? and then `emcpy.plots.plots.Scatter()`? If that doesn't work, what you have is obviously fine. Was just curious.

kevindougherty-noaa: I'll try that.

CoryMartin-NOAA: Thanks, I think that's clearer for the developers where the method comes from

CoryMartin-NOAA (approved):
Looks good to me, and all tests seem to pass. Will want @danholdaway to confirm things look good before merging in.

danholdaway (approved):

PR #62, opened 2022-11-07
added scripts to read gsi nc diags

kevindougherty-noaa:
## Description

Added script to read and filter through both conventional and satellite GSI netcdf4 diag files. Attempted to keep very similar structure of IODA script for consistency.


kevindougherty-noaa:
@CoryMartin-NOAA @danholdaway Tests failing due to lack of graphics information in GSI tests. Should I take the time to add this information or save that for a different PR?

CoryMartin-NOAA:
@kevindougherty-noaa what do you mean by 'graphics information'? We definitely want to include new unit tests with the new features.

src/eva/tests/config/testGsiObsSpaceAmsuaMetop-A.yaml
```diff
@@ -0,0 +1,21 @@
...
+  # Data read
+  # ---------
+- data:
+    type: GsiObsSpace
+    datasets:
+      - name: experiment
+        satellite: metop-a
+        sensor: amsua
+        filenames:
+          - /scratch2/NCEPDEV/stmp1/Kevin.Dougherty/ncDiags/gdas.20200922/00/atmos/diag_amsua_metop-a_ges.2020092200.nc4
```
danholdaway: These would need to be relative paths to the files downloaded by Git-LFS rather than on Hera.

kevindougherty-noaa: Yes, my fault. I made the change but didn't save it. Will fix that today.

kevindougherty-noaa:
Okay after several trial and error attempts and a brain blast this morning, I have figured out what was wrong. All tests pass and this is ready for review.

src/eva/data/gsi_obs_space.py
```diff
@@ -0,0 +1,244 @@
...
+    return next(g, True) and not next(g, False)
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+def uv(group_vars):
+    """
+    Adds 'uv' suffix to beginning of specified variables
+    if they are included input list of vars
```
CoryMartin-NOAA: missing an 'in'?

CoryMartin-NOAA (commented):
One comment that needs fixed + we need to subset the test data (use `ncks`) but this looks great!

src/eva/data/gsi_obs_space.py
```diff
@@ -0,0 +1,244 @@
...
+            if channels_str_or_list is not []:
+                channels = parse_channel_list(channels_str_or_list, self.logger)
+
+            # Filenames to be read into this collection
+            # -----------------------------------------
+            filenames = get(dataset, self.logger, 'filenames')
+
+            # File variable type
+            if 'satellite' in dataset:
+                satellite = get(dataset, self.logger, 'satellite')
```
CoryMartin-NOAA, line 169: we might need extra logic to handle ozone?

CoryMartin-NOAA (approved):
A minor comment but we can cross that bridge if we get to it.

danholdaway (approved):
Great! Will be nice to confirm that we get the same plots from a GSI file as an IODA file.

PR #64, opened 2023-01-13
Allow for numbers to be used in the arithmetic transform expression

danholdaway:
## Description

Allows numerals to be used in the arithmetic transform.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,7 @@ setuptools.setup(
         'cartopy>=0.18.0',
         'scikit-learn>=1.0.2',
         'xarray>=0.11.3',
-        'seaborn==0.12',
+#        'seaborn==0.12',
         'emcpy @ git+https://github.com/NOAA-EMC/' +
         'emcpy@4f36baf1a2302fb0daa49bd8415bb7d2a65347bb#egg=emcpy',
     ],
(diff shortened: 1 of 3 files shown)


src/eva/transforms/transform_utils.py
```diff
@@ -62,7 +62,7 @@ def replace_cgv(logger, collection, group, variable, *argv):
         else:
             arg_replaced.append(arg)
 
-    # Assert that the lenghts make sense
+    # Assert that the lengts make sense
```
CoryMartin-NOAA: still spelled wrong :-)

setup.py
```diff
@@ -39,6 +39,7 @@
         'cartopy>=0.18.0',
         'scikit-learn>=1.0.2',
         'xarray>=0.11.3',
+#        'seaborn==0.12',
```
CoryMartin-NOAA: I think PEP8 is complaining about this, perhaps the comment needs to be in front of the quote?

CoryMartin-NOAA (commented):
Two comments

src/eva/transforms/transform_utils.py
danholdaway: lol!

CoryMartin-NOAA (approved):
approved pending tests all pass

PR #67, opened 2023-01-24
IODA Version 3

danholdaway:
## Description

Update all the tests to IODA V3. Updates test files and YAMLs and source code where necessary. Where I found reliance on hardcoded names for channel or location outside of the IODA obs space class I tried to make it more generic.

Note that now the name for Location differs between the IODA and GSI classes so a new method is needed to change it to Location. I figure the default should be what it in IODA. The GSI class now has to call the method. In order to use the `sel` function in IODA you need to know the variable name ahead of time, hence why the Data class has to have a specific name.


src/eva/data/data_collections.py
```diff
@@ -230,6 +241,7 @@ def display_collections(self):
         self.logger.info('-'*80)
         self.logger.info(fcol.bold + 'Collections available: ' + fcol.end)
         for collection in self._collections.keys():
+            print(self._collections[collection])
```
CoryMartin-NOAA: is this for debugging or intended to stay?

CoryMartin-NOAA (commented):
minor question/comment

src/eva/data/data_collections.py
danholdaway: Good catch. Thanks.

danholdaway:
```suggestion
```

CoryMartin-NOAA (approved):

Issue #68, opened 2023-02-03
Add support for binary data files


EdwardSafford-NOAA:
## Description

*Add support for binary data files*

The eva package is going to be used to generate plots in the next generation of DA monitoring.  The current DA monitoring packages use binary files as an interim storage mechanism.  Extend eva to read these files and format the data into the correct xarray structures for eva image generation.  These files will be added to eva in order to accomplish this:

- eva/src/eva/data/mon_data_space.py  This will be similar to gsi_obs_space.py, but capable of handling binary data files and corresponding control files as input.
- eva/src/eva/tests/config/testMonDataSpaceHirs4Metop-A.yaml  A yaml file for testing mon_data_space.py.
- eva/src/eva/tests/data/time.hirs4_metop-a.ctl  A GrADS type control file describing the binary file format as currently produced by the RadMon DA monitor.
- eva/src/eva/tests/data/time.hirs4_metop-a.*.ieee_d  At least two binary data files as currently produced by the RadMon DA monitor.  More than one file will be necessary to confirm correct handling of multi-cycle data sets.

## Requirements

Given a yaml file with data inputs in binary format, produce data plots similar to those currently generated by the RadMon DA monitor.  

## Acceptance Criteria (Definition of Done)

Requirements above are met.

## Dependencies

None

EdwardSafford-NOAA:
@EdwardSafford-NOAA will work on this but I can't assign myself to it.  Maybe @CoryMartin-NOAA  can do that?

EdwardSafford-NOAA:
Progress to date:  Able to read control file and multiple binary data files (from RadMon) and load into a dataset.  Have figured out how to adjust dataset to fit with requested elements in yaml file including dropping vars and unrequested channels (a coordinate). 

Will next add the ability to drop specific regions.  (Note that "regions" here is something of a misnomer; they are actually surface types.  Region 1 is all, and that's primarily what's of interest.)

EdwardSafford-NOAA:
Closed with PR #70 .

CoryMartin-NOAA:
@EdwardSafford-NOAA best to not close the issue until the PR is merged, and putting `closes #68` in the PR will do so automatically upon merge!

EdwardSafford-NOAA:
Sorry, I must have missed something. I got email notification that PR #70 had been merged and saw this issue was still open.  Thought I was tidying up by closing this.  I was ignorant of the closes phasing for automatic closure; so noted.

[...]

PR #70, opened 2023-02-15
Add binary file read to support legacy DA extracted data files.

EdwardSafford-NOAA:
Add the ability to read the legacy DA monitors' binary data files.  

## Description
The next generation DA Monitoring is to be based on the eva package.  The plan is to enable monitor image generation from the legacy binary data files using eva.  Later the input source type will be swapped  when data extraction to NetCDF files is implemented.

This release allows eva to read RadMon binary files and create simple line plots.   A test yaml file and associated data files are included.

Additional changes are expected when reading other types of DA monitor binary files and production of additional monitoring graphics are undertaken.  This seems a reasonably sized start though.

This completes issue #68 .

@CoryMartin-NOAA @kevindougherty-noaa  @ADCollard


src/eva/data/mon_data_space.py
```diff
@@ -0,0 +1,302 @@
...
+                    self.logger.info(' inside len check ')
+                    requested_channels = parse_channel_list(str(channels_str_or_list), self.logger)
+                    drop_channels = True
+
+            # Get requested regions, convert to list
+            # --------------------------------------
+            regions_str_or_list = get(dataset, self.logger, 'regions', [])
+            self.logger.info(' regions_str_or_list = ' + str(regions_str_or_list))
+
+            #  Note:  use of regions_str_or_list is lifted directly from gsi_obs_space.py but it
```
EdwardSafford-NOAA: I think this is a bug.  I've documented the range of returned types in the code comments.  I note that if you try to run the testGsiObsSpaceAmsuaMetop-A.yaml with a single requested channel instead of 2 it fails, for the reason I've outlined.

src/eva/data/mon_data_space.py
```diff
@@ -0,0 +1,302 @@
...
+                        "total": (("channels", "regions"), total_sum_tmp),
+                        "omgbc": (("channels", "regions"), omgbc_sum_tmp),
+                        "omgnbc2": (("channels", "regions"), omgnbc_sum2_tmp),
+                        "total2": (("channels", "regions"), total_sum2_tmp),
+                        "omgbc2": (("channels", "regions"), omgbc_sum2_tmp),
+                    },
+                    coords={"channels": channo, "regions": np.arange(1,nregion+1)},
+                    attrs = {'satellite': satellite, 'sensor': sensor},
+                )
+                timestep_ds['times'] = cycle_tm
```
EdwardSafford-NOAA, line 109: I'm not sure I'm on the right track here.  The idea is to read in n data files, as specified in the testMonDataSpaceHirs4Metop-A.yaml file, and then concatenate the dataset arrays in individual_files by timestep.  That works fine, but I don't know how to then specify the time as the x axis when requesting a time series plot.  Time here is a coordinate not a variable and the plot fails for that reason.  I'll post the output of the final dataset below.  I'd be glad to learn what I'm doing wrong.

`INFO MonDataSpace: ds: <xarray.Dataset>`
`INFO MonDataSpace: Dimensions:   (times: 5, channels: 1, regions: 1)`
`INFO MonDataSpace: Coordinates:`
`INFO MonDataSpace:   * channels  (channels) int64 1`
`INFO MonDataSpace:   * regions   (regions) int64 1`
`INFO MonDataSpace:   * times     (times) int64 2015051418 2015051500 ... 2015051512 2015051518`
`INFO MonDataSpace: Data variables:`
`INFO MonDataSpace:     count     (times, channels, regions) float32 9.065e+03 ... 9.01e+03`
`INFO MonDataSpace: Attributes:`
`INFO MonDataSpace:     satellite:  metop-a`
`INFO MonDataSpace:     sensor:     hirs4`

EdwardSafford-NOAA: I tried a brute force hack to add a variable of cycle_time to the ds.  It sort of works; the counts are plausible, but the x label and increments make little sense:

hirs4_metop-a 0 count

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I just retested this and I'm convinced it is a bug.  If you edit `eva/src/eva/tests/config/testGsiObsSpaceAmsuaMetop-A.yaml` and change `channels: &channels 3,8` to `channels: &channels 3` this is the result:

`INFO EvaSetup: Starting Eva`
`INFO GsiObsSpace:   Initializing eva GsiObsSpace object`
`INFO EvaSetup: Running execute for GsiObsSpace`
`ABORT GsiObsSpace: In parse_channel_list the input is neither a list of integers or a string that can be parsed into a list of integers.`
`ABORT`

It's choking on 3 because it is not a list of integers nor is it a string; it's a single integer.

CoryMartin-NOAA: ah, yeah that is something we need to fix or else do `[3]` for in the meantime.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Now with a datetime object instead it looks much better:

hirs4_metop-a 0 count

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I'll open a new issue for it then.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Now with a datetime object instead it looks much better:

hirs4_metop-a 0 count

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Issue #71 was opened to handle this and PR #72 has been submitted for the fix.

src/eva/tests/config/testMonDataSpaceHirs4Metop-A.yaml
```diff
@@ -0,0 +1,50 @@
...
+      - ../data/time.hirs4_metop-a.ctl
+    datasets:
+      - name: experiment
+        satellite: metop-a
+        sensor: hirs4
+        ctlfile:
+#          - ${data_input_path}/diag_amsua_metop-a_ges.2020092200.nc4
+          - ../data/time.hirs4_metop-a.ctl
+        filenames:
+          - ../data/time.hirs4_metop-a.2015051418.ieee_d
```
danholdaway: Is there a reason you can't use the absolute path by keeping `${data_input_path}`?

danholdaway:
You can run `python pycodestyle_run.py` in the top directory to see the failures on your machine.

src/eva/tests/config/testMonDataSpaceHirs4Metop-A.yaml
EdwardSafford-NOAA: That will use ${data_input_path} in the final, non-draft version.

For my own edification, where to I need to set ${data_input_path} when running my tests?  That wasn't immediately clear so I just used this quick hack to get it running.

EdwardSafford-NOAA:
@danholdaway I had not noticed that before.  Super useful - thanks!

src/eva/data/mon_data_space.py
```diff
@@ -0,0 +1,283 @@
...
+
+class MonDataSpace(EvaBase):
+
+    # ----------------------------------------------------------------------------------------------
+
+    def execute(self, data_collections, timing):
+
+        # Get name of the control file to parse
+        # -------------------------------------
+        ctlfile = self.config.get('ctlfile')
```
CoryMartin-NOAA: perhaps change to `control file` to be more explicit/clear?

src/eva/data/mon_data_space.py
```diff
@@ -0,0 +1,283 @@
...
+        mystr = ""
+        for x in ctlfile:
+            mystr += ' ' + x
```
CoryMartin-NOAA: It's not clear to me what the point of these 3 lines are

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Done.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Me either; they have been removed.

src/eva/data/mon_data_space.py
```diff
@@ -0,0 +1,280 @@
...
+        for dataset in self.config.get('datasets'):
+
+
+            # Set the collection name
+            # -----------------------
+            collection_name = get(dataset, self.logger, 'name')
+
+            # Get control file and parse
+            # --------------------------
+            control_file = get(dataset, self.logger, 'control_file')
```
CoryMartin-NOAA, line 43: FYI this is fine, but there is no need for `snake_case`, you can use `control file` in the YAML and it is fine

EdwardSafford-NOAA: That makes sense.  Force of habit in the yaml file.

Note that I figured out that the control file was in the wrong location in the yaml file.  It was initially an element (if that's the right term) of `data` instead of `dataset`.  I've corrected that.

CoryMartin-NOAA:
@EdwardSafford-NOAA looks like some coding norms issues with the latest commit(s) but this looks mostly good to me.

kevindougherty-noaa (approved):
Looks good to me. There's a lot of interest in these capabilities so excited to implement some of these features elsewhere.

PR #73, opened 2023-02-27
Eva Interactive

asewnath:
## Description

New EvaInteractive module provides users with a jupyter notebook-friendly interface to Eva.


.github/workflows/eva_tests_notebook.yml
```diff
@@ -0,0 +1,53 @@
...
+
+    # Install eva
+    - name: Upgrade pip
+      run: $CONDA/bin/pip3 install --upgrade pip
+    - name: Install eva and dependencies
+      run: $CONDA/bin/pip3 install --use-deprecated=legacy-resolver -r requirements-github.txt --user .
+    - name: Put eva in the path
+      run: echo "$HOME/.local/bin" >> $GITHUB_PATH
+
+    # Install hvplot
```
danholdaway: You might be able to avoid needing this by adding it to `requirements-github.txt`

src/eva/data/eva_interactive.py
```diff
@@ -0,0 +1,118 @@
...
+from eva.utilities.timing import Timing
+from eva.eva_base import EvaFactory
+from eva.transforms.arithmetic import arithmetic
+from eva.transforms.accept_where import accept_where
+# --------------------------------------------------------------------------------------------------
+
+
+class EvaInteractive():
+
+    def __init__(self, filename):
```
danholdaway: Can 'filename' actually be an argument to `load_data_collection`? That might be more future proof if eventually the collection could contain more than more files worth of data.

src/eva/data/eva_interactive.py
```diff
@@ -0,0 +1,118 @@
...
+        self.data_collection = DataCollections()
+        self.logger = Logger('EvaInteractive')
+        self.collection = 'interactive'
+        self.timer = Timing()
+        self.filename = filename
+        self.var_cache = []
+
+    def load_data_collection(self):
```
danholdaway: Can you rename this to `load_ioda` or something like that? Since it's hardcoded underneath.

src/eva/eva_tests.py
```diff
@@ -60,6 +61,44 @@ def application_tests(logger):
...
+
+        # Open notebook using nbconvert
+        with open(os.path.join(eva_path, 'tests', 'notebooks', notebook)) as file:
+            nb_in = nbformat.read(file, nbformat.NO_CONVERT)
+
+        # Replace variables in notebook
+        updated_nb_in = replace_vars_notebook(nb_in, **overwrite_dict)
+
+        # Execute notebook
+        ep = ExecutePreprocessor(timeout=600, kernel_name='python3')
```
danholdaway, line 95: Nice find!

danholdaway (commented):
Thanks Akira! The changes look good to me. Great to start building these notebook features into eva.

danholdaway:
I just noticed also that the actual notebook file is quite large at 7mb. That will increase the size of the repo quite a lot over time. I don't quite see why since the hs plots are not embedded in there. Do you see why it's so big. We'll need to use GitLFS to track these if they are to be this large.

CoryMartin-NOAA (approved):
looks good to me once the conflict is resolved

PR #74, opened 2023-02-27
Fix Location coordinate values to make compatible with Skylab

danholdaway:
## Description

When skylab merges ioda files the Location coordinate in the file ends up with something a bit unusual. Only the values from the first file are present and the rest are zero. This is mightily confusing for xarray and bypasses the logic we had in eva for checking if Location coordinate is set properly, because it was assuming zeros everywhere if not set. 

This code fixes that by forcing the Location coordinate to index values. In testing the Eva code with the very large (1.4million obs) ABI GOES-16 file we also noticed the read part was very slow. This fix also eliminates this slowness.

The attached plot of 1,440,392 observations is created in around 21 seconds (instead of minutes before) with the read step taking only 2.15 seconds.

observations_abi_g16_brightnessTemperature_8

Yaml file: 

``` YAML
diagnostics:

  # Data read
  # ---------
- data:
    type: IodaObsSpace
    datasets:
      - name: experiment
        missing_value_threshold: 400.0
        filenames:
          - fb.abi_g16_bt_8km.2022-02-15T00:00:00Z_0000.nc4
        channels: &channels 8
        groups:
          - name: ObsValue
            variables: &variables [brightnessTemperature]
          - name: hofx
          - name: EffectiveQC
          - name: MetaData

  graphics:

    # Map plots
    # ---------

    # Observations
    - batch figure:
        variables: *variables
        channels: *channels
      dynamic options:
        - type: vminvmaxcmap
          channel: ${channel}
          data variable: experiment::ObsValue::${variable}
      figure:
        figure size: [20,10]
        layout: [1,1]
        title: 'Observations | ABI GOES 16 | Obs Value'
        output name: map_plots/abi_g16/${variable}/${channel}/observations_abi_g16_${variable}_${channel}.png
      plots:
        - mapping:
            projection: plcarr
            domain: global
          add_map_features: ['coastline']
          add_colorbar:
            label: ObsValue
          add_grid:
          layers:
          - type: MapScatter
            longitude:
              variable: experiment::MetaData::longitude
            latitude:
              variable: experiment::MetaData::latitude
            data:
              variable: experiment::ObsValue::${variable}
              channel: ${channel}
            markersize: 2
            label: ObsValue
            colorbar: true
            cmap: ${dynamic_cmap}
            vmin: ${dynamic_vmin}
            vmax: ${dynamic_vmax}
```

## Dependencies

N/A

## Impact

N/A

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -43,16 +43,6 @@ def subset_channels(ds, channels, add_channels_variable=False):
 # --------------------------------------------------------------------------------------------------
 
 
-def check_location(location):
-    if max(location) == 0:
-        new_location = range(location.size)
-        location = new_location + location
-    return location
-
-
-# --------------------------------------------------------------------------------------------------
-
-
 class IodaObsSpace(EvaBase):
 
     # ----------------------------------------------------------------------------------------------
@@ -86,6 +76,7 @@ class IodaObsSpace(EvaBase):
 
             # Loop over filenames
             # -------------------
+            total_loc = 0
             for filename in filenames:
 
                 # Assert that file exists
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):

BenjaminRuston:
@danholdaway thanks,, and yeesh, what an oddity

Issue #75, opened 2023-03-02
[Bug] Indentation issue in gsi_obs_space.py


kevindougherty-noaa:
## Current behavior (describe the bug)

@ADCollard identified an issue when he tries to run multiple files using gsi_obs_space.py, he only gets information from the last file in the list.

There should be an indentation on this line here: https://github.com/JCSDA-internal/eva/blob/develop/src/eva/data/gsi_obs_space.py#L238

The `data_collections.create_or_add_to_collection(collection_name, ds, 'nobs')` is outside of the filename loop, which is causing this issue.

danholdaway:
Thanks @kevindougherty-noaa are you going to tackle this? Perhaps we can improve one of the IODA tests to ingest multiple obs files? Say AMSUA-A from 00z and 06z. You could probably cheat and have two copied of the 00z named differently.

kevindougherty-noaa:
@danholdaway I assigned @ADCollard as he said he would not mind tackling it. However, I can as well.

ADCollard:
@danholdaway @kevindougherty-noaa I plan to address this.  Hopefully today.

PR #78, opened 2023-03-06
Fix indent to allow multiple GSI diag files to be read in

ADCollard:
The indent of the 
```
data_collections.create_or_add_to_collection(collection_name, ds, 'nobs')
```
line in `src/eva/data/gsi_obs_space.py` means that only one file (the last one listed) is read in.

This PR fixes the indent.

References issue [#75](https://github.com/JCSDA-internal/eva/issues/75)

### DIFF
--- a/src/eva/data/gsi_obs_space.py
+++ b/src/eva/data/gsi_obs_space.py
@@ -234,8 +234,8 @@ class GsiObsSpace(EvaBase):
                                           group_name + '\' in file ' + filename +
                                           ' does not have any variables.')
 
-            # Add the dataset to the collections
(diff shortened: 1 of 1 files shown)


ADCollard:
A sample map with data from four input files!

Screenshot 2023-03-06 at 1 59 04 PM

CoryMartin-NOAA (approved):

kevindougherty-noaa (approved):

danholdaway:
Should we adjust the testing to make sure this is protected? I.e. plot two files in the test?

ADCollard:
> 

@danholdaway As far as I can see, there is no numerical test being done anyway.   As this fails silently do we need to add this?

Issue #77, opened 2023-03-06
Location of MapScatter in map_scatter.py is wrong [Bug]


EdwardSafford-NOAA:
File `diagnostics/map_scatter.py` has the wrong location for `emcpy.plots.map_plots.MapScatter`.  Instead it uses `emcpy.plots.plots.MapScatter` which fails.

## To Reproduce

Run `eva testIodaObsSpaceAmsuaN19.yaml`.  It will fail in` eva/diagnostics/map_gridded.py` at line 18 when if tries to create a `plotobj` using `emcpy.plots.plots.MapGridded`.

## Expected behavior
Test case should run to completion.

## Additional information (optional)
I'll be glad to fix this if I can be assigned to it.

Changing the location to `emcpy.plots.map_plots.MapScatter` works -- it looks like `emcpy` changed at some point and `eva `didn't get updated.  Unless there's a reason to import all of `map_plots `I propose importing only `emcpy/plots/map_plots/MapScatter`  in `map_gridded.py` and not using the explicit path to `MapScatter `in the code.

danholdaway:
Thanks for finding and being willing to fix this @EdwardSafford-NOAA

EdwardSafford-NOAA:
@danholdaway
I'm running into a situation I don't understand.  In `diagnostics/map_scatter.py` the problem is simply that `emcpy.plots.plots` needs to be changed to `emcpy.plots.map_plots`.  Making that change allows the test case to run to completion.  But if I change the import line to 

`from emcpy.plots.map_plots import MapScatter`

and then try to use MapScatter directly 

`self.plotobj = MapScatter(latvar, lonvar, datavar)` 

the result is a seemingly unrelated error codition:
`Traceback (most recent call last):`
  `File "/home/Edward.Safford/.local/bin/eva", line 8, in <module>`
    `sys.exit(main())`
  `File "/home/Edward.Safford/.local/lib/python3.9/site-packages/eva/eva_base.py", line 226, in main`
    `eva(config_file)`
  `File "/home/Edward.Safford/.local/lib/python3.9/site-packages/eva/eva_base.py", line 202, in eva`
   ` eva_figure_object.execute(data_collections, timing)`
  `File "/home/Edward.Safford/.local/lib/python3.9/site-packages/eva/plot_tools/figure_driver.py", line 95, in execute`
    `self.make_figure(figure_conf_fill, plots_conf_fill,`
[...]

PR #79, opened 2023-03-07
Update hash for emcpy; fix mapscatter syntax

CoryMartin-NOAA:
## Description

This PR closes issue #77 

@EdwardSafford-NOAA found that there is a syntax issue with EMCpy now if using a more recent version in EVA. This PR bring EVA up to date with EMCpy.

## Dependencies

None

## Impact

Tagging @danholdaway as he may want to use a new EVA tag for SWELL soon...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setuptools.setup(
         'xarray>=0.11.3',
         #  'seaborn==0.12',
         'emcpy @ git+https://github.com/NOAA-EMC/' +
-        'emcpy@4f36baf1a2302fb0daa49bd8415bb7d2a65347bb#egg=emcpy',
+        'emcpy@dcdc868e7f48bb1ff55507f34c0548742af8bec1#egg=emcpy',
         # 'seaborn==0.12',
     ],
     package_data={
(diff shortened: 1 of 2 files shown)


CoryMartin-NOAA:
@EdwardSafford-NOAA I can't make you a reviewer because I don't have admin privileges and you aren't added to this repo 'officially'. But could you test and report results back here?

EdwardSafford-NOAA:
@CoryMartin-NOAA With these changes `test/testIodaObsSpaceAmsuaN19.yaml` runs to completion.

As you pointed out earlier the change that was made to `map_scatter.py` also needs to be made to `map_gridded.py`.   I may have overlooked it but I haven't found a test case that uses map_gridded.py, so I don't think there is an easy way to test that.

danholdaway (approved):

danholdaway:
I've merged to get things working again. But I thought we were referring to a specific commit of EMCpy. How could that commit being referenced change and the eva tests have passed in whatever PR did that?

CoryMartin-NOAA:
I am not sure, but I noticed the issue on Orion. Perhaps when you pip install it doesn't update the commit automatically if it already is installed? I think in the CI everything stayed working, but updating local existing installs led to the discrepancy.

CoryMartin-NOAA:
@danholdaway actually seems the more likely scenario is "user error" aka me, I think I installed the update eva first, then installed emcpy, which defaulted to develop in the virtual environment...

danholdaway:
OK. Thanks for the info.

Issue #76, opened 2023-03-02
Add avg and sdv calculations across a given time span


EdwardSafford-NOAA:
## Description
For DA Monitoring 2.0 we need the ability to calculate avg and std values in a dataset for a given variable across a specified time span.   Typically these time spans are 1 day (4 cycles) and 30 days (180 cycles) and are plotted by channel (x axis).

Add necessary function(s) in `utilities/stats.py` to accomplish this and extend `data/mon_data_space.py` as necessary.

@EdwardSafford-NOAA  will complete this work.

## Requirements

Given inputs of a dataset containing data from multiple cycles, a variable in the dataset, and a range, return datasets containing the average and standard deviation values for the variable.  The range is to start with the latest cycle in the input dataset.

Additionally `data/mon_data_space.py` will need to be extended to parse yaml instructions to create these avg and sdv values and then create a line plot with those values by channel (or other specified variable). 

## Acceptance Criteria (Definition of Done)

Given a yaml file containing data files for a range of cycle times and instructions for plotting the avg and sdv values of a variable (i.e. obs count) for a time span that includes some and all of the cycle times (2 cases), create a line plot with the values arranged by channel.

## Dependencies

None

EdwardSafford-NOAA:
@CoryMartin-NOAA  or @kevindougherty-noaa  can you assign this to me?  Thanks.

EdwardSafford-NOAA:
Progress update.

I'm working on constructing summary plots for the monitoring.  They include both single cycle plots (typically last 4 cycles) and averages (4 cycles, 28 cycles, 120 cycles).  I'm starting plotting the last 4 cycles together.   I've figured out what I need is most a transform operation as contained in `src/eva/transform`.  It doesn't look like what I need is fully implemented, but it's hard to tell since I'm really just learning both xarray and the mapping from yaml to `src/eva/transform`.  So I'm hard coding my way through the xarray manipulation within `src/eva/data/mon_data_space` to get what I need working and then I'm going to map that to the capabilities in transform and extend the transform capabilities as needed.

I'm reading in multiple binary files and paring down the dataset to the requested vars from the yaml, which is just obs count to keep it simple.  What I need to do is select the count variable for the first 4 cycles and merge that into the dataset as a new variable so it can be plotted.  What I'm doing is:

`ds_cnt = ds[['GsiIeee::count']]`
`ds_cnt1 = ds_cnt.sel(times='2015051418')`
[...]

PR #83, opened 2023-03-24
Add select_time transform

EdwardSafford-NOAA:
This PR adds a `select_time` data transform.  This provides means to select a subset of a variable by individual cycle or by time slice.  The selected subset is added to the data_collections as directed by the transform.

Additionally some changes have been made to `data/mon_data_space.py` to standardize it with respect to the other data loading files and common xarray standards.  The new `config/testMonSummary.yaml` file uses the `select_time` transform to manipulate data and then generate images similar to RadMon summary plots including plots of mean values for specific multi-cycle periods.

## Dependencies
None.

## Impact
None.

Ref #76

### DIFF
--- a/src/eva/transforms/select_time.py
+++ b/src/eva/transforms/select_time.py
@@ -1,0 +1,90 @@
+# (C) Copyright 2021-2022 NOAA/NWS/EMC
+#
+# (C) Copyright 2021-2022 United States Government as represented by the Administrator of the
+# National Aeronautics and Space Administration. All Rights Reserved.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+from eva.transforms.transform_utils import parse_for_dict, split_collectiongroupvariable
+from eva.transforms.transform_utils import replace_cgv
+from eva.utilities.config import get
+from eva.utilities.logger import Logger
+
+
+# --------------------------------------------------------------------------------------------------
+#  Select a variable by the Time dimension in form YYYYMMDDHH.  If the input is a single time
+#  then values for the specified variable will be selected for that cycle and added to
+#  data_collections.
+#
+#  If two cycle times are included they will be used as the bounds for a time slice, and the
+#  average of that span will be added to the data_collections.  Note that when specifying a slice
+#  the order must be older (lower) to newer (higher).
+
+def select_time(config, data_collections):
+
+    # Create a logger
+    logger = Logger('SelectTimeTransform')
+
+    # Parse config for dictionary
+    [collections, groups, variables] = parse_for_dict(config, logger)
+
+    # Parse config for the expression and new collection/group/variable naming
+    new_name_template = get(config, logger, 'new name')
+    starting_field_template = get(config, logger, 'starting field')
+
+    # Get cycles from yaml file and confirm there are 1 or 2 times
+    cyc = get(config, logger, 'cycles')
+    if type(cyc) is list:
+        cycles = cyc
+    else:
+        cycles = [cyc]
+
+    if None in cycles:
+        logger.abort('cycles: for transformation not specified in yaml file. ' +
+                     'This should be cycles: followed by 1 or 2 cycle times ' +
+                     'in YYYYMMDDHH format.')
+    if len(cycles) > 2:
+        logger.info('WARNING:  more than 2 cycles specified in yaml file. ' +
+                    'Only the first 2 times will be used.')
+
+    # Loop over the templates
+    for collection in collections:
+
+        for group in groups:
+
+            for variable in variables:
+
+                if variable is not None:
+                    # Replace collection, group, variable in template
+                    [new_name, starting_field] = replace_cgv(logger, collection, group, variable,
+                                                             new_name_template,
+                                                             starting_field_template)
+                else:
+                    new_name = new_name_template
+                    starting_field = starting_field_template
+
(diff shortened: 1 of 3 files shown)


EdwardSafford-NOAA:
The testing failures appear due to change made to the cycle variable in `MonDataSpace.py` and when cycle is then used as a plot variable. This fix may take a little time so I've moved this PR to draft.

EdwardSafford-NOAA:
@CoryMartin-NOAA @danholdaway @kevindougherty-noaa  Please take a look when you have a moment.  I would have assigned you as reviewers but I don't have permission to do that.

src/eva/transforms/select_time.py
```diff
@@ -0,0 +1,90 @@
...
+    # Parse config for the expression and new collection/group/variable naming
+    new_name_template = get(config, logger, 'new name')
+    starting_field_template = get(config, logger, 'starting field')
+
+    # Get cycles from yaml file and confirm there are 1 or 2 times
+    cyc = get(config, logger, 'cycles')
+    if type(cyc) is list:
+        cycles = cyc
+    else:
+        cycles = [cyc]
```
kevindougherty-noaa: `cycles = cyc if type(cyc) is list else [cyc]`

kevindougherty-noaa (commented):
Looks good to me, just one minor comment for more concise conditional.

src/eva/transforms/select_time.py
EdwardSafford-NOAA: Cool.  Done.

src/eva/transforms/select_time.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Parse config for dictionary
+    [collections, groups, variables] = parse_for_dict(config, logger)
+
+    # Parse config for the expression and new collection/group/variable naming
+    new_name_template = get(config, logger, 'new name')
+    starting_field_template = get(config, logger, 'starting field')
+
+    # Get cycles from yaml file and confirm there are 1 or 2 times
+    cyc = get(config, logger, 'cycles')
```
CoryMartin-NOAA: when I first read the YAML, when I saw `cycles` I thought you would specify a list of all the cycles, but it seems it is the start/end bound.

What are your thoughts on having `cycle` be specified if you want to read just one cycle, and then otherwise use keys like `start cycle` and `end cycle` (and perhaps a third key `window length`?) to make it clear that it is a range?

EdwardSafford-NOAA: That would be fine.  So the yaml parsing would try `cycle` first, and if not found would try `start cycle` and `end cycle`, right?  

I'm not sure `window length` would work in all cases -- occasionally there are missing cycles in the data and `cycle interval` would have to be specified as well.  My thinking so far is that the supporting Mon 2.0 infrastructure would determine which cycles are available and generate the corresponding yaml files for image generation.

CoryMartin-NOAA: Yes exactly, check for `cycle` first, then if not that, look for `start cycle` and `end cycle`. And okay we can punt on window length, that works for me.

EdwardSafford-NOAA: Very good.  I'll make it so.

EdwardSafford-NOAA: I don't see a way in `get` (in `config.py`) to continue if a key isn't found rather than `sys.exit`.  Unless I missed an established solution I'm going to modify `get` to add an `abort_on_failure` parameter (default True).

CoryMartin-NOAA: I think you can do:
```
cyc = get(config, logger, 'cycle')
if not cyc:
    start_cyc = get(config. logger, 'start cycle')
...

if not cyc and not start_cyc and not end_cyc:
   sys.exit()
```
(pseudo code but do you get what I"m saying?)

CoryMartin-NOAA: perhaps I'm misreading your comment (sorry), can you add an optional extra argument that is the default which is empty or something like that?

EdwardSafford-NOAA: I think we're on the same page.  The problem is if `get `in this line
`cyc = get(config, logger, 'cycle')` 
doesn't find `cycle`, `get `fires `logger.abort`.  I'm going to add a parameter to `get` to call/not call `logger.abort` with the default set to the current behavior (call `logger.abort`).

CoryMartin-NOAA: Yeah that sounds good, while it does allow for a default to be passed, I think this is just one of many cases where we want 'None' to be a valid option.

EdwardSafford-NOAA: Changes have been made.    Please see what you think.

src/eva/utilities/config.py
```diff
@@ -56,15 +56,15 @@ def get(self, key, default=None):
...
+def get(dict, logger, key, default=None, abort_on_failure=True):
 
     if default is None:
 
         if key in dict:
 
             return dict.get(key)
 
-        else:
+        elif abort_on_failure:
```
CoryMartin-NOAA, line 67: it may be good to have a warning print that the key is missing and the default is used, but I won't ask that be done here unless others want that.

CoryMartin-NOAA (approved):
Looks good besides my one comment, but I want to make sure @danholdaway has a chance to look at this before it gets merged in, in case he has a better suggestion on how to handle this use case.

src/eva/utilities/config.py
kevindougherty-noaa: I always think it is good to inform the user of things like this.

EdwardSafford-NOAA: I'll go ahead and add a warning then.

Issue #85, opened 2023-04-06
Modify mon_obs_space.py to make more generic


EdwardSafford-NOAA:
## Description

`data/mon_obs_space.py` was recently added to make legacy RadMon time-series (binary) data files available as a data source for the EVA package.  The code from this initial effort explicitly names the data file components in this way:

```
timestep_ds = Dataset(
                    {
                        "count": (("channels", "regions"), count_tmp),
                        "penalty": (("channels", "regions"), penalty_tmp),
                        "omgnbc": (("channels", "regions"), omgnbc_sum_tmp),
                        "total": (("channels", "regions"), total_sum_tmp),
                        "omgbc": (("channels", "regions"), omgbc_sum_tmp),
                        "omgnbc2": (("channels", "regions"), omgnbc_sum2_tmp),
                        "total2": (("channels", "regions"), total_sum2_tmp),
                        "omgbc2": (("channels", "regions"), omgbc_sum2_tmp),
                        "cycle": (("channels", "regions"), cycle_tmp),
                    },
```

That works, but it is specifically tailored to the RadMon time series data.  There are additional types of legacy DA binary files which will need to be supported as part of the Mon 2.0 effort with different variables and dimension names.

Accordingly make `data/mon_obs_space.py` more generic.  Start with `get_ctl_stats()`, which parses and returns information from a control file.  Instead of returning explicitly named variables, return dicts for the dimensions, variables, and attributes.  Then use that information to load the file contents in a generalized fashion.  Once this is working for the RadMon time series data it should be simple to add additional monitor and data file types, starting with the RadMon scan-angle data files.

Please assign this to @EdwardSafford-NOAA 

## Requirements

Load data specified in the corresponding yaml file.  There should be no difference in plotted results from running  `eva testMonDataSpaceHirs4Metop-A.yaml` with the new (generic) version of `data/mon_obs_space.py` and the results from running the same with the current develop version.   

Additionally new test data for RadMon scan angle data files will be included in the release along with a test yaml file to generate image plots.  These should run to completion.

## Dependencies

None

EdwardSafford-NOAA:
I've run into some complications with the RadMon angle binary files.  They are 3-D not 2-D like the time binary files, and they have some eccentricities.  It's a solvable problem but it adds enough complexity that doing this in 2 pieces will make code reviews easier. 

[...]

PR #86, opened 2023-04-17
Make mon_data_space.py more generic (part 1)

EdwardSafford-NOAA:
## Description

This PR is the first step to making the file loads for the binary legacy DA monitor data files generic.  

The first cut at `data/mon_data_space.py` was able to load only one specific type of monitor files (RadMon, time-series).  This PR accomplishes the same file load into `data_collections` but does so in a generic way that will support additional DA monitors (which all use binary files but with different dimensions).

Note that in many places there is an if/elif/else on the number of dimensions in use. This will be 1, 2, or 3.  2 works with these changes; 1 and 3 are not fully formed yet.  The changes in this PR work with the existing test cases `testMonDataSpaceHirs4Metop-A.yaml` and `testMonSummary.yaml`.  I opted to halt work at this point and complete the work for dimensions 1 and 3 in follow-on release(s) so as to keep the changes brief enough for an effective code review.

## Dependencies

None

## Impact

None

This PR is part 1 for changes in #85.


EdwardSafford-NOAA:
@CoryMartin-NOAA or @kevindougherty-noaa please assign reviewers to this PR.  Also a question -- it looks like the Application Test fails with the `testIodaObsSpaceAircraft.yaml` case.  Is that a known problem or did I just step in something?

CoryMartin-NOAA:
@EdwardSafford-NOAA not sure why that is failing, looks like it is trying to download shapefiles... but odd it would start doing that now. I don't think it is something you did, we may need to investigate separately

EdwardSafford-NOAA:
@CoryMartin-NOAA for what it's worth I was able to download that shape file (to my local machine), and it appears to be ok.  I wonder if the test's download just got interrupted.  I'd re-run the tests but I don't find a way to do that (I'm guessing that's another permission thing).

CoryMartin-NOAA:
@EdwardSafford-NOAA good suggestion let me try that, yeah my guess is you have to be a 'maintainer' or whatever permission I have on this repo

EdwardSafford-NOAA:
@CoryMartin-NOAA well, heck.  Seemed like a good idea, but it failed in the same place.

CoryMartin-NOAA:
@EdwardSafford-NOAA I can confirm the same issue is happening re-running tests on develop... so this needs to be investigated. I'm helping organize a code sprint this week so not sure how much time I'll have to diagnose, will get back to you soon once I/others can find time

EdwardSafford-NOAA:
FWIW I ran `testIodaObsSpaceAircraft.yaml` with both the develop and my PR branches.  Both run to completion without error.

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
 
-        count = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        penalty = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        omgnbc_sum = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        total_sum = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        omgbc_sum = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        omgnbc_sum2 = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        total_sum2 = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
-        omgbc_sum2 = f.read_reals(dtype=np.dtype('>f4')).reshape(nchans, nregion)
+        if ndims_used == 1:
```
CoryMartin-NOAA, line 300: Can we at least combine these somehow where the if statement defines the empty size and dims that get reshaped? the for loops should (mostly, maybe need a dim=3 version) be able to be written only once

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+
+    # ----------------------------------------------------------------------------------------------
+
+    def load_dset(self, vars, nvars, coords, darr, dims, ndims_used,
+                  x_range, y_range, z_range, cyc_darr):
+
+        # create dataset from file components
+        rtn_ds = None
+
+        if ndims_used == 1:
```
CoryMartin-NOAA: Similar comment to above, I think these may be able to be consolidated a bit, but 4:15pm Friday brain doesn't have the exact solution

src/eva/utilities/utils.py
```diff
@@ -336,3 +336,13 @@ def slice_var_from_str(config, datavar, logger):
 
 
 # --------------------------------------------------------------------------------------------------
+
+
+def is_number(s):
```
CoryMartin-NOAA, line 341: does this exist already in python somewhere maybe? or not?

EdwardSafford-NOAA: Seems like it should but I haven't found it so far.  This solution was a cut and paste from stackoverflow.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Fair point.  I've made it a bit more concise.  The dim=3 case is still a special case, but maybe I can figure a way to make it less special.

src/eva/utilities/utils.py
CoryMartin-NOAA: works for me, just wanted to check

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+        cycstrs = file_name.split('.')
 
         for cycstr in cycstrs:
             if cycstr.isnumeric():
                 cycle_tm = datetime(int(cycstr[0:4]), int(cycstr[4:6]),
                                     int(cycstr[6:8]), int(cycstr[8:]))
 
+        filename = os.path.join(file_path, file_name) if file_path else file_name
+
+        if not os.path.isfile(filename):
```
CoryMartin-NOAA, line 292: What about printing a warning (or error) about missing filename using the logger?

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I've got an idea here.  I think if I create a dictionary from the variable dimensions and coordinates I can then create the dataset via `xr.Dataset.from_dict(d)`.  That would put the if `num_dims` logic inside the for loop.  Giving it a shot now.

src/eva/data/mon_data_space.py
CoryMartin-NOAA: Sounds like a plan, this is already more concise but if you can make it even better with your new plan, that would be great!

CoryMartin-NOAA (commented):
One more comment, but after your next round I think this will be close to merging, thanks for the hard work so far!

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+        # - "Channel" can be either the x or y coordinate and can be
+        #      numbered non-consecutively, which has been captured in channo.
+        # - The z coordinate is never used for channel.
+
+        if dims['xdef'] > 1:
+
+            if coords['xdef'] == 'Channel':
+                x_range = channo
+            else:
+                x_range = np.arange(1, dims['xdef']+1)
```
kevindougherty-noaa: Logic is fine, a more concise way:
`x_range = channo if coords['xdef'] == 'Channel' else np.arange(1, dims['xdef']+1)`

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+
+        if dims['ydef'] > 1:
+            if coords['ydef'] == 'Channel':
+                y_range = channo
+            else:
+                y_range = np.arange(1, dims['ydef']+1)
```
kevindougherty-noaa: Same as above:
`y_range = channo if coords['ydef'] == 'Channel' else np.arange(1, dims['ydef']+1)`

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,279 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+        if ndims_used == 3:
+            cycle_arr = np.reshape([[cycle_tm] * dims['xdef'] * dims['ydef'] * dims['zdef']],
+                                   (dims['xdef'], dims['ydef'], dims['zdef']))
+        elif ndims_used == 2:
+            cycle_arr = np.reshape([[cycle_tm] * dims['xdef']] * dims['ydef'],
+                                   (dims['xdef'], dims['ydef']))
+        elif ndims_used == 1:
+            cycle_arr = np.reshape([[cycle_tm] * dims['xdef']], (dims['xdef']))
+        else:
+            self.logger.abort(f'ndims_used must be in range of 1-3, value is {ndims_used}')
```
kevindougherty-noaa: Could possibly use a dictionary like factory here:

```
d = {
1: np.reshape([[cycle_tm] * dims['xdef']], (dims['xdef'])),
2: np.reshape([[cycle_tm] * dims['xdef']] * dims['ydef'],
                      (dims['xdef'], dims['ydef'])),
3: np.reshape([[cycle_tm] * dims['xdef'] * dims['ydef'] * dims['zdef']],
                       (dims['xdef'], dims['ydef'], dims['zdef']))
}

try:
    cycle_arr = d[ndims_used]
except KeyError:
    self.logger.abort(f'ndims_used must be in range of 1-3, value is {ndims_used}')
```

I have not tested this, but could be a solution.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I'll add a warning.  It's not uncommon for there to be missing files over a given time span.  Whatever scripting solution I/we use to create the yaml files should step over the gaps and give the yaml a clean list, but the bottom line is one missing file isn't a fatal condition.

kevindougherty-noaa (commented):
Just a few suggestions I saw. Nothing that absolutely needs to be done, just could help add concise code.

src/eva/data/mon_data_space.py
CoryMartin-NOAA: Ok that's fine but yes something like `logging.info(f"File {filename} is missing")` should be sufficient

EdwardSafford-NOAA: Done.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: See if you like this solution.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I like it and have made it so.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Done.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: It is a solution.  Cool!  It's about the same code length but I went this way so I'd remember the construct for future reference.

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,256 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
 
         f.close()
+        return rtn_array, cycle_tm
+
+    # ----------------------------------------------------------------------------------------------
+
+    def cycle_to_np_array(self, dims, ndims_used, cycle_tm):
+
+        # build numpy array with requested dimensions
+        d = {
```
CoryMartin-NOAA, line 346: this is nice!

src/eva/data/mon_data_space.py
CoryMartin-NOAA: looks good!

CoryMartin-NOAA (approved):
Thanks @EdwardSafford-NOAA ! I'll let @kevindougherty-noaa do another pass through but once he gives the :+1: I'll merge this in.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Super!  Please note that this PR won't close #85.  It's step 1 of probably 2 PRs.  Thanks.

src/eva/data/mon_data_space.py
```diff
@@ -200,87 +196,256 @@ def subset_coordinate(self, ds, coordinate, requested_subset):
...
+                {
+                    'cycle': ((coords['xdef'], coords['ydef'], coords['zdef']), cyc_darr),
+                },
+                coords={coords['xdef']: np.arange(1, dims['xdef']+1),
+                        coords['ydef']: np.arange(1, dims['ydef']+1),
+                        coords['zdef']: np.arange(1, dims['zdef']+1)},
+            )
+
+        rtn_ds = rtn_ds.merge(new_cyc)
+        return rtn_ds
```
kevindougherty-noaa: Your logic is not wrong at all here, but I think I found a way to make this a bit more efficient. I used the factory-like method similar to what I suggested last review. It gets rid of all of the conditionals which are slow in Python. Please double check to make sure this works:

```
ndims_dict = {
    1: {
        'dims': (coords['xdef']),
        'cycle': Dataset({'cycle': ((coords['xdef']), cyc_darr),},
                         coords={coords['xdef']: np.arange(1, dims['xdef']+1)},)
    },
    2: {
        'dims': (coords['xdef'], coords['ydef']),
        'cycle': Dataset({'cycle': ((coords['xdef'], coords['ydef']), cyc_darr),},
                         coords={coords['xdef']: np.arange(1, dims['xdef']+1),
                         coords['ydef']: np.arange(1, dims['ydef']+1)},)
    },
    3: {
        'dims': (coords['xdef'], coords['ydef'], coords['zdef']),
        'cycle': Dataset({'cycle': ((coords['xdef'], coords['ydef'], coords['zdef']), cyc_darr),},
                         coords={coords['xdef']: np.arange(1, dims['xdef']+1),
                         coords['ydef']: np.arange(1, dims['ydef']+1),
                         coords['zdef']: np.arange(1, dims['zdef']+1)},)
    }
}

# Create empty dict
d = {}

# Loop through vars and fill empty dict
for x in range(0, nvars):
    d[vars[x]] = {
        'dims': ndims_dict[ndims_used]['dims'],
        'data': darr[x, :, :]
    }

# Create Dataset from dictionary
rtn_ds = Dataset.from_dict(d)

# Merge cycle Dataset based on ndims_used
rtn_ds = rtn_ds.merge(ndims_dict[ndims_used]['cycle'])
```

kevindougherty-noaa (commented):
I added a suggestion that may save some lines of code and helps get rid of multiple conditionals.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Cool.  I had tried doing that here before but, looking at your code, I see where I might have gone wrong in that attempt.  I'll give it another go.

EdwardSafford-NOAA: @kevindougherty-noaa I'm getting this error with the 'cycle' definition in the factory-like construct:

`ValueError: Could not convert tuple of form (dims, data[, attrs, 
encoding]): ('Channel', array([[[datetime.datetime(2023, 4, 9, 0, 0), .... to Variable`

Any thoughts on how to correct that?

kevindougherty-noaa: @EdwardSafford-NOAA I am not too sure what you mean. Is it in issue in the dictionary or when you try and merge?

EdwardSafford-NOAA: It's in the dictionary.  I've tried paring down/simplifying the dictionary structure.  The 'dims' definition is fine.  The 'cycle' is the source of the problem.

kevindougherty-noaa: Maybe `Dataset()` can't work in the dictionary. I unfortunately don't have a solution for the error you are getting.

EdwardSafford-NOAA: I worked my way around the `Dataset `issue.  This isn't quite as concise as what you'd hoped for but all the the `if/elif/else `on `num_dims `is now eliminated.  Please see what you think.

kevindougherty-noaa (approved):
Looks great. I appreciate your patience and willingness to try the suggested changes!

EdwardSafford-NOAA:
Thanks very much for your suggestions.  It's always great to learn new things!

PR #88, opened 2023-04-19
Feature/fv3 obs space

kevindougherty-noaa:
## Description

Adds a new needed obs space script for plotting 6-tile FV3 model files. Includes changes to existing files as well as new added files.


danholdaway:
I'm not sure this is the correct name for the class. Rather than ModelObsSpace perhaps it should be FV3Restart?

kevindougherty-noaa:
@danholdaway Thank you, this was a temporary name because I was not able to get any name with FV3 in it because when it calls the appropriate script based on the class name, the number 3 causes it to fail. @CoryMartin-NOAA just suggested to name it CubedSphere or something along those lines.

danholdaway:
For whatever name you choose note that we will end up adding another class that reads cubed sphere history, both GMAO and NCEP history. So we'll want to be able to distinguish between Restarts, as you have here, and History. 

Are you planning to add a new transform that interpolates the cube to a generic latlon grid that can be plotted with cartopy etc?

CoryMartin-NOAA:
@danholdaway I think we went with `CubedSphereRestart` for this class now. In terms of transforms, this plots on a map with Cartopy already, so I don't think there is a need to convert to LatLon. @kevindougherty-noaa should have an example somewhere here that he can upload that shows what the plots will look like with EVA.

src/eva/data/cubed_sphere_restart.py
```diff
@@ -0,0 +1,122 @@
+# (C) Copyright 2021-2022 NOAA/NWS/EMC
```
CoryMartin-NOAA: 2023!

src/eva/data/cubed_sphere_restart.py
```diff
@@ -0,0 +1,122 @@
...
+                ds = xr.Dataset(var_dict)
+
+                # Assert that the collection contains at least one variable
+                if not ds.keys():
+                    self.logger.abort('Collection \'' + dataset['name'] + '\', group \'' +
+                                      group_name + '\' in file ' + filename +
+                                      ' does not have any variables.')
+
+            # Add the dataset to the collections
+            data_collections.create_or_add_to_collection(collection_name, ds, 'nobs')
```
CoryMartin-NOAA: 'nobs'?

src/eva/data/cubed_sphere_restart.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # Nan out unphysical values
+        data_collections.nan_float_values_outside_threshold(threshold)
+
+        # Change the channel dimension name
+        data_collections.adjust_channel_dimension_name('nchans')
```
CoryMartin-NOAA: shouldn't be any channels here...

src/eva/tests/config/testCubedSphereRestart.yaml
```diff
@@ -0,0 +1,66 @@
...
+  # Data read
+  # ---------
+- data:
+    type: CubedSphereRestart
+    datasets:
+      - name: experiment
+        variable: T
+        resolution: C768
+        fv3_filenames:
+          - ${data_input_path}/20210801.000000.sfc_data.tile1.nc
```
CoryMartin-NOAA: hmm, we need to figure out how to download this data for the test, and also go lower resolution... the operational resolution is a bit much for this purpose...

src/eva/tests/config/testCubedSphereRestart.yaml
```diff
@@ -0,0 +1,66 @@
...
+          - ${data_input_path}/20210801.000000.sfc_data.tile6.nc
+        orog_filenames:
+          - ${data_input_path}/C768_oro_data.tile1.nc
+          - ${data_input_path}/C768_oro_data.tile2.nc
+          - ${data_input_path}/C768_oro_data.tile3.nc
+          - ${data_input_path}/C768_oro_data.tile4.nc
+          - ${data_input_path}/C768_oro_data.tile5.nc
+          - ${data_input_path}/C768_oro_data.tile6.nc
+        groups:
+          - name: FV3NcDiag
```
CoryMartin-NOAA: NcDiag is not an appropriate group name here I don't think, go with `FV3Restart` maybe?

danholdaway:
Thanks for the info @CoryMartin-NOAA. I didn't realize that Cartopy could plot a cubed sphere gridded field. Good to know.

kevindougherty-noaa:
image

@danholdaway here is the output from the files in the yaml file.

danholdaway:
Nice! Looks good.

src/eva/data/cubed_sphere_restart.py
```diff
@@ -1,6 +1,6 @@
-# (C) Copyright 2021-2022 NOAA/NWS/EMC
+# (C) Copyright 20212-2023 NOAA/NWS/EMC
```
CoryMartin-NOAA: Typo :-)

CoryMartin-NOAA (commented):
Tiny typo, then ready to merge

CoryMartin-NOAA (approved):
Thanks @kevindougherty-noaa looks great!

PR #89, opened 2023-04-20
Fix application test errors in CI

CoryMartin-NOAA:
## Description

CI stopped working properly for the EVA application tests. No idea why it stopped working but updating to the latest Cartopy and Seaborn seems to have fixed it.

### DIFF
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -5,7 +5,7 @@ matplotlib==3.5.2
 cartopy==0.21.1
 scikit-learn==1.1.2
 xarray==2022.6.0
-seaborn>=0.12
+seaborn==0.12.2
 hvplot==0.8.2
 nbconvert==6.5.4
 bokeh==2.4.3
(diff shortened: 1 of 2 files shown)


requirements-github.txt
```diff
@@ -2,10 +2,10 @@ pyyaml==6.0
 pycodestyle==2.9.1
 netCDF4==1.6.1
 matplotlib==3.5.2
-cartopy==0.21.0
+cartopy==0.21.1
 scikit-learn==1.1.2
 xarray==2022.6.0
-seaborn==0.12
+seaborn>=0.12
```
danholdaway: More things that can change underneath us with non-exact versions. Can you check which version GH cloned and then set to that?

danholdaway:
Thanks for fixing this! Seems to happen every three months or so, not ideal

requirements-github.txt
CoryMartin-NOAA: yeah this was just a matter of throwing things at the wall and see what stuck, will update to that, test, and then make this non-draft

danholdaway (approved):

PR #96, opened 2023-05-12
Add Channel_Number to MetaData group and fix Cluster issue

ADCollard:
Two issues were found when trying to plot channel stats for IASI when the input data to UFO came from the IODA converters.

1) The 'Cluster' dimension comprises all zeros and was causing the code to crash.
2) The sensorChannelNumber arrays had dimensions of channels * locations and was inconsistent with channel_stats (it had dimension of number of channels when using obs files created from GSI).

These are fixed by:
1) Forcing 'Cluster' to have definite values: [1,2,3,4,5,6,7]
2) Creating a new variable within the MetaData group that is the 1D channel number.

### DIFF
--- a/src/eva/data/ioda_obs_space.py
+++ b/src/eva/data/ioda_obs_space.py
@@ -22,13 +22,13 @@ import netCDF4 as nc
 
 def subset_channels(ds, channels, add_channels_variable=False):
 
-    if 'channel' in list(ds.dims):
+    if 'Channel' in list(ds.dims):
 
         # Number of user requested channels
         channel_use = len(channels)
 
         # Number of channels in the file
-        channel_in_file = ds.channel.size
+        channel_in_file = ds.Channel.size
 
         # If user provided no channels then use all channels
         if channel_use == 0:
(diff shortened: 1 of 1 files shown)


src/eva/data/ioda_obs_space.py
```diff
@@ -22,13 +22,13 @@
 
 def subset_channels(ds, channels, add_channels_variable=False):
 
-    if 'channel' in list(ds.dims):
+    if 'Channel' in list(ds.dims):
```
CoryMartin-NOAA, line 25: I don't know why this was lower case before, but I think it still has to be, as the CI tests fail now for not being able to find 'channel'

ADCollard: @CoryMartin-NOAA I am a little confused by that warning message as I thought this referred to the dimensions read in to the data space.  [ioda_obs_space.amsua_n19.hofx.2020-12-14T210000Z](https://github.com/JCSDA-internal/eva/blob/develop/src/eva/tests/data/ioda_obs_space.amsua_n19.hofx.2020-12-14T210000Z.nc4) is using upper case `Channel`.  Also that warning message shows up when running the current version.

```
Hera(hfe09):/scratch1/NCEPDEV/da/Andrew.Collard/git/eva/src/eva/tests/data$ ncdump -h ioda_obs_space.amsua_n19.hofx.2020-12-14T210000Z.nc4
netcdf ioda_obs_space.amsua_n19.hofx.2020-12-14T210000Z {
dimensions:
	Channel = 15 ;
	Location = UNLIMITED ; // (100 currently)
variables:
	int Channel(Channel) ;
		Channel:_FillValue = -2147483647 ;
	int Location(Location) ;
		Location:_FillValue = -2147483647 ;

// global attributes:
		string :NCO = "netCDF Operators version 5.0.1 (Homepage = http://nco.sf.net, Code = http://github.com/nco/nco)" ;
		string :_ioda_layout = "ObsGroup" ;
		:_ioda_layout_version = 0 ;
		string :converter = "gsi_ncdiag.py" ;
```

CoryMartin-NOAA: this isn't a warning though, it's failing:
```
KeyError: "'channel' is not a valid dimension or coordinate"
```

ADCollard: Sorry, I was looking in the last test.  It is failing in `testCubedSphereRestart.yaml`.  What is that about?  I will start digging.

PR #98, opened 2023-05-12
A few random bugfixes 

CoryMartin-NOAA:
1. status badges were pointing to @danholdaway 's repo and not JCSDA-internal
2. MonDataSpace test worked in CI but not on Hera due to the way it determined the file's valid time
3. tests will now only run if it is a .yaml file
4. a warning message that was added is a red herring and is not actually correct, so I'm removing it

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,10 +5,10 @@
 
 | Test      | Status  |
 | --------- | --------|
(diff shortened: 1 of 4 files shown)


danholdaway (approved):

ADCollard (approved):
Looks good to me

PR #99, opened 2023-05-16
Make DA monitor handling more generic (part 2)

EdwardSafford-NOAA:
## Description

These changes improve the handling of DA monitor data.  The dimensions and coordinates in use (as specified in the DA monitor control file) are now dynamic lists instead of string literals.  Parsing these lists is now done by dictionaries instead of `if|elif|else` syntax.

Tests have been successfully completed using RadMon time files (2 dimensions), RadMon angle files (3 dimensions), and OznMon time files (1 or 2 dimensions).  

Please don't close issue #85 yet though.  I think there still may be some work to do to handle batch plots.  I'd like to get this reviewed and merged before exploring that (potential) complication.

## Dependencies
None

## Impact
None

Ref #85


src/eva/utilities/config.py
```diff
@@ -36,20 +36,20 @@ def __init__(self, dict_or_yaml, logger):
...
                 return super().get(key)
 
-            else:
+            elif abort_on_failure:
 
                 self.logger.abort("Configuration does not have the key")
 
         else:
-
+            logger.info(f'Warning: key {key} not found, returning default {default}')
```
CoryMartin-NOAA: I removed this in #98 as the message is not correct.
This will warn even if the key is found as it is currently written.

CoryMartin-NOAA (commented):
One small comment, but otherwise I think this looks good, but want a 2nd pair of eyes to take a look.

src/eva/utilities/config.py
EdwardSafford-NOAA: Ok, I've deleted the output.

CoryMartin-NOAA (approved):
Looks good, but will wait for another approval/our chat this afternoon before merging

EdwardSafford-NOAA:
Hmmm.  Not sure why the Notebook tests have now decided to fail:

`ImportError: cannot import name 'ImportedStyleSheet' from 'bokeh.models' (/home/runner/.local/lib/python3.10/site-packages/bokeh/models/__init__.py)`

CoryMartin-NOAA:
@EdwardSafford-NOAA you're not alone here, this is also in #102 , so it seems like a GitHub actions issue.

CoryMartin-NOAA:
Note, don't worry about the Eva Notebook Tests failure, that will be fixed in #102

ADCollard (approved):
Looks good to me.

PR #105, opened 2023-05-25
Make CubeSphereRestart handle 3D fields

CoryMartin-NOAA:
## Description

CubeSphereHistory data ingest worked great, but only for 2D fields (like in the test). This makes a few changes to make it able to handle reading 3D fields.


kevindougherty-noaa (approved):
Looks good to me. May want to include a 3D example case, but I understand the issues that arise with storing that kind of data on the repo.

PR #108, opened 2023-06-20
Add support for MinMon data

EdwardSafford-NOAA:
## Description

Extend `mon_data_space.py` to support MinMon data sources, which are 1 dimensional arrays, and have no header or footer (as Fortran binary files do).  Add `Iteration` as an available data dimension.

Add a mechanism to `mon_data_space.py` to zero out data for missing files.  Time series data occasionally has missing data files and in these cases the desired plot is a cycle time with zero/missing data. 

Lastly extend `transforms/arithmetic.py` by adding log() transform.  This is needed by MinMon gnorm plots.

## Dependencies

None

## Impact

None


EdwardSafford-NOAA:
@CoryMartin-NOAA @kevindougherty-noaa @ADCollard please review if you have a moment.  Thanks.

src/eva/data/mon_data_space.py
```diff
@@ -291,7 +292,8 @@ def read_ieee(self, file_name, coords, dims, ndims_used, dims_arr, nvars, vars,
 
         # find cycle time in filename and create cycle_tm as datetime object
         cycle_tm = None
-        cycstrs = file_name.split('.')
+        fname = file_name.replace('/', '.')
+        cycstrs = fname.split('.')
```
kevindougherty-noaa: If you don't need the variable `fname` anywhere else, you can combine these two lines into:
`cycstrs = file_name.replace('/', '.').split('.')`

However, it doesn't need to be changed.

kevindougherty-noaa (approved):
Looks good to me. Just had a minor comment that doesn't need to be changed.

src/eva/data/mon_data_space.py
```diff
@@ -300,52 +302,66 @@ def read_ieee(self, file_name, coords, dims, ndims_used, dims_arr, nvars, vars,
 
         filename = os.path.join(file_path, file_name) if file_path else file_name
 
+        load_data = True
         if not os.path.isfile(filename):
```
CoryMartin-NOAA: nit pick, but I would prefer the `if` be flipped and not do an `if not` and then `else` but rather `if` then `else`

src/eva/data/mon_data_space.py
```diff
@@ -300,52 +302,66 @@ def read_ieee(self, file_name, coords, dims, ndims_used, dims_arr, nvars, vars,
...
-        if ndims_used == 1:
-            rtn_array = np.empty((0, dims[dims_arr[0]]), float)
+        if ndims_used == 1:		# MinMon
+            rtn_array = np.empty((0, dims[dims_arr[0]]), dtype=float)
+            if not load_data:
+                zarray = np.zeros((dims[dims_arr[0]]), float)
             dimensions = [dims[dims_arr[0]]]
 
-        if ndims_used == 2:
+        if ndims_used == 2:		# RadMon time, bcoef, bcor, OznMon time
```
CoryMartin-NOAA, line 317: this might break pycodestyle conventions?

src/eva/data/mon_data_space.py
```diff
@@ -300,52 +302,66 @@ def read_ieee(self, file_name, coords, dims, ndims_used, dims_arr, nvars, vars,
...
             rtn_array = np.empty((0, dims[dims_arr[0]], dims[dims_arr[1]]), float)
+            if not load_data:
+                zarray = np.zeros((dims[dims_arr[0]], dims[dims_arr[1]]), float)
             dimensions = [dims[dims_arr[0]], dims[dims_arr[1]]]
 
-        if ndims_used == 3:
+        if ndims_used == 3:		# RadMon angle
```
CoryMartin-NOAA, line 323: same comment, I think that might be too many spaces (unless the linter doesn't complain then it's fine)

CoryMartin-NOAA (commented):
Couple of minor things

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: No apparently not.  All tests passed.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: As above linter is good with this.

src/eva/data/mon_data_space.py
CoryMartin-NOAA: Ok, works for me.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: No problem.  Change made.

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: Fair point, change made.

CoryMartin-NOAA (approved):
Looks good, thanks!

PR #109, opened 2023-06-21
Eva Refactoring

asewnath:
## Description

Refactoring eva for more clarity.

List of changes:
- Decoupled eva base class from drivers. The eva base class and factory is reserved for runtime generation of data handlers. We plan to eventually add another abstractmethod to eva base so that data handlers generate their own default config for eva interactive (this means also removing config from the eva base constructor).
- The main 'eva' method is moved to eva_driver.py 
- Drivers are turned into functions. Logger and timer are provided to the functions
- setup.py and eva_tests are updated to reflect these changes
- 'diagnostics' and 'data' are removed from all yaml files
- updated files to not expect 'diagnostics' and 'data' keywords


src/eva/eva_driver.py
```diff
@@ -0,0 +1,90 @@
+# (C) Copyright 2021-2022 NOAA/NWS/EMC
```
CoryMartin-NOAA: I know the copyrights are stupid, but probably should update to 2023 here.

danholdaway (approved):

src/eva/eva_driver.py
CoryMartin-NOAA: But don't not merge based on this comment.

CoryMartin-NOAA (approved):
Looks good, thanks for doing this!

PR #111, opened 2023-06-21
eva_driver fix

Dooruk:
I wasn't able to execute `eva` from the CLI without the missing modules: `argparse`, `os`


asewnath:
Thanks for this fix, looks like I also need to test this using the command line instead of just relying on the github actions!

asewnath (approved):

PR #114, opened 2023-06-26
Feature/soca data class

Dooruk:
SOCA data class ("inspired" by CubeRestart class) and test files for 5 degrees map plot. Works with SOCA analysis outputs (initiated with restart or history files). Tripolar grid coordinates are obtained from `soca_gridspec.nc`. Reads 2d & 3d fields and able to plot slices using this [utility](https://github.com/JCSDA-internal/eva/blob/develop/src/eva/utilities/utils.py#L327). This didn't break 2d field variable plotting as demonstrated in the test so I didn't separate 2d and 3d variables.

Handles ocn, ice, wav, and sfc variables (i.e., SOCA outputs), as they are all on the same grid.

2/3 degrees examples with SOCA history file input :

## **SSH**
soca_2dfield_midres_ave_ssh
## **SSS**
soca_2dfield_midres_Salt


CoryMartin-NOAA:
FYI @AndrewEichmann-NOAA and @guillaumevernieres

CoryMartin-NOAA (approved):
Looks great! Thanks for adding this

src/eva/tests/config/testSocaRestart.yaml
```diff
@@ -0,0 +1,41 @@
+# Data read
+# ---------
+datasets:
+  - name: experiment
+    type: SocaRestart
+    soca_filenames: ${data_input_path}/ocn.3dvar_soca.an.2018-04-15T00:00:00Z.nc
+    geometry_file: ${data_input_path}/soca_gridspec.72x35x25.nc
+
+    variables: [ave_ssh, Salt]
+    coordinate variables: [lon, lat]
```
danholdaway, line 10: Does the name of these ever change? Perhaps they can be hardcoded?

Dooruk: `soca_gridspec` is resolution dependent, so no. There is scattered grid in `soca_gridspec` as well (lonu, latv). I thought they might be utilized if need be?

danholdaway: Would the plotting as it's written work if you chose something other than lon, lat? I.e. could you pick lonu,latu and plot Salt with it?

Issue #117, opened 2023-06-27
Add horizontal and vertical line plots 


EdwardSafford-NOAA:
## Description

Eva does not currently have an interface to use emcpy's HorizontalLine and VerticalLine classes.  Add these in eva/src/eva/diagnostics.

## Requirements

None

## Acceptance Criteria (Definition of Done)

A yaml file can produce a plot that includes a horizontal and vertical line at the specified y and x coordinates respectively.

## Dependencies

None

EdwardSafford-NOAA:
I will be glad to do this if I can be assigned.  Thanks.

EdwardSafford-NOAA:
Closed with PR #118 .

PR #118, opened 2023-06-29
Add HorizontalLine and VerticalLine to diagnostics

EdwardSafford-NOAA:
## Description

Added HorizontalLine and VerticalLine classes to src/diagnostics.  The second plot has 2 horizontal and 2 vertical lines plotted using these new classes.

image

## Dependencies

None

## Impact

None

Completes #117

### DIFF
--- a/src/eva/diagnostics/horizontal_line.py
+++ b/src/eva/diagnostics/horizontal_line.py
@@ -1,0 +1,28 @@
+from eva.eva_path import return_eva_path
+from eva.utilities.utils import get_schema, update_object
+import emcpy.plots.plots
+import os
+
+
+class HorizontalLine():
+
+    def __init__(self, config, logger, dataobj):
+
+        # Get the y value to plot
+        # -----------------------
+        yval = config['y']
+
+        # Create declarative plotting HorizontalLine object
+        # -------------------------------------------
+        self.plotobj = emcpy.plots.plots.HorizontalLine(yval)
+
+        # Get defaults from schema
+        # ------------------------
+        layer_schema = config.get('schema', os.path.join(return_eva_path(), 'defaults',
+                                  'horizontal_line.yaml'))
+        config = get_schema(layer_schema, config, logger)
+#        delvars = ['y', 'type', 'schema']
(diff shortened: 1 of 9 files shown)


src/eva/diagnostics/horizontal_line.py
```diff
@@ -0,0 +1,28 @@
...
+        # Create declarative plotting HorizontalLine object
+        # -------------------------------------------
+        self.plotobj = emcpy.plots.plots.HorizontalLine(yval)
+
+        # Get defaults from schema
+        # ------------------------
+        layer_schema = config.get('schema', os.path.join(return_eva_path(), 'defaults',
+                                  'horizontal_line.yaml'))
+        config = get_schema(layer_schema, config, logger)
+#        delvars = ['y', 'type', 'schema']
```
CoryMartin-NOAA: remove comment

CoryMartin-NOAA (commented):
One very minor thing. Also can we add one or both of these to one of the test YAMLs so that it gets exercised?

src/eva/diagnostics/horizontal_line.py
EdwardSafford-NOAA: Comment removed.   Is adding lines to an existing plot sufficient or would you prefer a stand-alone example in the test yamls (yamlae? yamlii? what is the correct plural for yaml?)?

CoryMartin-NOAA: no need for a new test, just throw something in one of the YAMLs. Like add a line at 0 on a histogram or something.

EdwardSafford-NOAA: Done.  The result of `testIodaObsSpaceAmsuaN19.yaml`'s density plots is so:

image

CoryMartin-NOAA (approved):

src/eva/diagnostics/horizontal_line.py
CoryMartin-NOAA: Looks good, thanks!!

danholdaway:
Thanks for adding this

Issue #120, opened 2023-07-05
[Bug] Problem with channels in mon_data_space


EdwardSafford-NOAA:
## Current behavior (describe the bug)

Channels for `mon_data_space` objects are defined in the input control file.  Channel numbers are in ascending order, but they are not guaranteed to be continuous nor must they start at 1.  Testing with abi_g16 data has revealed problems handling channels in some cases.  Abi_g16 uses 10 channels, numbered 7-16.  The channels in the abi_g16 summary plots appear as 1-10 but with data for only 4 channels in the first two plots.  

The channel specification in the control (*.ctl) files specifies both the number of channels listed as 1-n, and the actual channel numbers, which for abi_g16 are 7-16.  There is a problem in the interaction between these two values. 

## To Reproduce

Run eva with `/scratch1/NCEPDEV/da/Edward.Safford/noscrub/git/eva/src/eva/tests/config/monSummary.yaml`.  I'm running on hera but this isn't platform specific. The results are thus:
![image](https://github.com/JCSDA-internal/eva/assets/62339196/75c40d61-115e-4222-b528-9540ae870f8a)

## Expected behavior

The summary plots should show channels 7-16.  Data values should be indicated for all channels, channels 8 and 10 should be shown as assimilated, while all others should be shown as non-assimilated.

## Additional information (optional)

I'll work this (and have already started) if I may be assigned to it.

EdwardSafford-NOAA:
Found the problem, but it wasn't actually with channels.  There was a simple problem with loading from column major to row major for binary files with data in 2-d arrays.  I'll clean things up and issue a PR tomorrow.

![image](https://github.com/JCSDA-internal/eva/assets/62339196/8bc39efa-ec64-4e7b-9f03-c3928a61526e)

EdwardSafford-NOAA:
Closed with PR #125

PR #125, opened 2023-07-11
Feature/fix chan 120

EdwardSafford-NOAA:
## Description

The problems with non-continuous channels turned out to be a data load issue instead.  My row-major brain hadn't quite figured out the correct dimension order for reading the binary files.  Fixing that fixed the channel assignment issues, and the summary plots now are in very close agreement with the legacy RadMon summary plots (see below).  I had thought the earlier discrepancies in the eva plots might be due to incorrectly purging missing data but the problem was more fundamental. 

Eva plot for abi_g16 at 2023070218:
image

Legacy RadMon summary plot for abi_g16 at 2302307:
image

The eva plot is using 'tight layout' and it makes a difference.  Also worth noting, multi-line titles work but the yaml must use " not ' for strings in order to recognize the newline (\n) character.

I added two additional horizontal lines to optimally position the channel markers.

## Dependencies

None

## Impact

None

Completes #120

### DIFF
--- a/src/eva/data/mon_data_space.py
+++ b/src/eva/data/mon_data_space.py
@@ -129,7 +129,8 @@ class MonDataSpace(EvaBase):
                 ds['chan_assim'] = (['Channel'], chan_assim)
                 ds['chan_nassim'] = (['Channel'], chan_nassim)
                 ds['chan_yaxis_100'] = (['Channel'], [-100]*len(channo))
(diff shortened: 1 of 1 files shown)


ADCollard (approved):

EdwardSafford-NOAA:
Closes #120

PR #126, opened 2023-07-13
Bump to v1.5.0

danholdaway:
## Description

With recent YAML changes we need a minor version update.


CoryMartin-NOAA (approved):
Bumped!

PR #127, opened 2023-07-13
Plotting/ Eva Interactive Refactoring

asewnath:
## Description

- Changed the directory structure so that `diagnostics`, `defaults`, and `plot_tools` are located under `plotting/emcpy`.
- Added the `plotting/hvplot` directory and migrated all eva interactive hvplot routines there
- Changed files to reflect new directory structure
- Refactored eva interactive transform and plotting functions to act as client interface
- EvaBase renamed to EvaDatasetBase. EvaDatasetBase no longer requires a config during initialization. This parent class also contains a new virtual method for generating a default config. Every data space will be required to define a default config for eva interactive.


danholdaway (approved):
Thanks Akira. I think this structuring of the code is much clearer, where the various plotting backends are represented side by side and with an obvious place to add future backends.

danholdaway:
If Cory agrees with this structuring we can merge and then tag as 1.5.0

src/eva/eva_interactive.py
```diff
@@ -0,0 +1,170 @@
...
+        eva_object = creator.create_eva_object(eva_class_name, 'data', self.logger, self.timer)
+        config = eva_object.generate_default_config(filenames, collection_name)
+        print(config)
+        eva_object.execute(config, data_collection, self.timer)
+
+        self.dc_dict[collection_name] = data_collection
+        self.fn_dict[collection_name] = filenames[0]
+
+        # Open up file to find channel requirements
+        if eva_class_name is not 'JediLog':
```
CoryMartin-NOAA: I think this also needs to be for `mon_data_space` but we don't need to support that datatype here if you don't want

CoryMartin-NOAA (approved):
This all makes sense to me, two parallel backends for plotting. Thanks for doing all of this!

src/eva/eva_interactive.py
asewnath: Definitely will want to support mon_data_space in interactive. I'll make a note of this. Right now MonDataSpace has an empty generate_default_config method because I wasn't sure how to populate the config for it.

CoryMartin-NOAA: @EdwardSafford-NOAA can you help Akira with what the config for MonDataSpace should contain?

EdwardSafford-NOAA: Sure, I'd be glad to help.  Does that need to happen now (in this PR) or as a follow-up item?  I'll have to study the code a bit to understand what the config should include for `MonDataSpace`.

EdwardSafford-NOAA: As a first cut I'd include these items in the default_config:
```
'satellite' : None,
'sensor': None,
'control_file': control_file,
'filenames': [ ],
'channels' : [ ],
'regions': [ ],
'levels': [ ],
'groups': [ ],
'name': collection_name
```

asewnath:
@kevindougherty-noaa @EdwardSafford-NOAA Just a note - it seems that the application test was failing because an issue regarding "tight layout" when using matplotlib 3.7.1 (https://github.com/JCSDA-internal/eva/actions/runs/5612753990/jobs/10271031666). For now I've pinned matplotlib to 3.5.2 and everything is working as expected.

PR #136, opened 2023-08-11
Add eva-docs dry run workflow

asewnath:
Create workflow to call eva-docs dry run build to make sure the website still builds with proposed eva changes. The associated eva-docs dry run workflow is here: https://github.com/JCSDA-internal/eva-docs/blob/develop/.github/workflows/docs_dry_run.yaml


asewnath:
@danholdaway the pr dry run is finally working as expected

danholdaway (approved):

PR #153, opened 2023-08-24
Add level components, streamline level, channel, and scan components

### LINKED ISSUE #152: Add support for Levels as per Channels to mon_data_space.py
## Description

The OznMon plots use level much in the same way as the RadMon uses channels.  To that end level, level_assim, and level_nonassim variables need to be added to the dataset to support plots by level and assimilated/non-assimilated level markers to  summary plots.

## Dependencies

None

EdwardSafford-NOAA:
Add level components in a manner similar to channels.  Convert channel information into dictionary (like levels), streamline handling of conditional variables related to channel, level, and scan position.

## Dependencies

None

## Impact

None

Closes #152


EdwardSafford-NOAA:
This is an example of the levels in use in an OznMon summary plot:

summary ompsnp_npp

EdwardSafford-NOAA:
I see the documentation failure but I don't understand what's going wrong.  The `Details` are not overly detailed.  Thoughts?

CoryMartin-NOAA:
@asewnath any clues about the EVA documentation test failure? Seems unrelated to this PR

src/eva/data/mon_data_space.py
```diff
@@ -689,56 +699,51 @@ def load_dset(self, vars, nvars, coords, darr, dims, ndims_used,
...
-            scan_array[:, x] = np.array([scanpo])
-            dataset['scan'] = (['Scan', 'Channel'], scan_array)
+        if chans_dict is not None:
+            dataset['channel'] = (['Channel'], chans_dict["chan_nums"])
+            if len(chans_dict["chans_assim"]) > 0:
+                dataset['chan_assim'] = (['Channel'], chans_dict["chans_assim"])
+            if len(chans_dict["chans_nassim"]) > 0:
+                dataset['chan_nassim'] = (['Channel'], chans_dict["chans_nassim"])
+
+            dataset['chan_yaxis_100'] = (['Channel'], [-100]*len(chans_dict["chan_nums"]))
```
CoryMartin-NOAA: This is for plotting, right? Is there a way to do this on the plotting/YAML side rather than making pseudo data? If not, then this is fine.

EdwardSafford-NOAA: Yes its used as the y axis to position the channel and level markers.  I tried using a fixed y value (-1.5) in the yaml but that produced an error.  Should it be able to accept numeric values for x or y axes?

I did think of a slightly better way to handle the assimilated/non-assimilated data for channel and level markers.  I'll open a new issue to implement that.

CoryMartin-NOAA: I guess I'm wondering if this is a 'transform' that can multiply a variable by X to get a new variable?

EdwardSafford-NOAA: Hmmm.  That's an interesting idea.  I think I'd need a single variable of correct dimensions with all zero values, and then in the yaml I could apply an arithmetic transform to make it what I need.   That's a much better way of handling it.  Thanks!

That's enough of a change that I think I'd like to take that up in a new issue.

EdwardSafford-NOAA: Let me walk part of that back. I just did a quick test and it works.  I'll make that change here.  I'll move this PR to draft and switch it back when I've got the change implemented and tested.

CoryMartin-NOAA (approved):
Looks good, thanks!

CoryMartin-NOAA:
@EdwardSafford-NOAA your changes look fine, but I can't merge until we figure out the EVA documentation issue (which I am 99% sure is irrelevant to your PR)

asewnath:
Looking into this

asewnath:
Can we try retriggering the tests now? I don't see immediately see the branch for this PR, otherwise I would do an empty commit

CoryMartin-NOAA:
@asewnath looks like it's fixed, thanks!

PR #155, opened 2023-08-25
Eva docs dry run test

asewnath:
Patching to make eva dry run build work while the eva branch problem is being sorted out.


asewnath:
The eva-docs workflow update takes care of the failing test, but if you look at the logs the build can't locate eva. This is probably fixed by updating conf.py to include the eva/src in path, but this might have unintended consequences for the actual documentation build so more investigating needs to be done.

Issue #156, opened 2023-09-19
Add support for batch processing by level 


EdwardSafford-NOAA:
## Description

Extend the batch processing mechanism to plot data by channels to instead plot by levels.  This is needed in order to generate the Ozone and Conventional DA monitoring plots which use level as a coordinate instead of channel.

I'm not sure if/how this might relate to #124 so I've opened it as a separate issue.  

## Requirements

Given an Ozone time series yaml file, generate plots by level.

## Dependencies

None

EdwardSafford-NOAA:
Closed by #157 .

Issue #124, opened 2023-07-10
Add support for batch level/layer processing for figures


CoryMartin-NOAA:
## Description

Like the title says, use the batch processing capability to process multiple (all) model level/layers for figure generation.

PR #157, opened 2023-09-20
Expand batch plots to support processing by level

EdwardSafford-NOAA:
## Description

Expand batch plots to handle levels like channels.  Step through levels, creating data subsets by level, and plotting each level.  This is an example of the output (note the holes in the data are actually missing data files and are not errors):
ompsnp_npp 1 omg

@CoryMartin-NOAA I'm not sure if this partially answers #124.

## Dependencies

None

## Impact

None

Completes #156 .


src/eva/plotting/emcpy/plot_tools/figure_driver.py
```diff
@@ -65,9 +65,21 @@ def figure_driver(config, data_collections, timing, logger):
...
             channels_str_or_list = batch_conf.get('channels', [])
             channels = parse_channel_list(channels_str_or_list, logger)
+            step_vars = channels
+            step_var_name = 'channel'
+            title_fill = ' Ch. '
+
+            # Get list of levels
+            levels_str_or_list = batch_conf.get('levels', [])
+            levels = parse_channel_list(levels_str_or_list, logger)
+            if levels != []:
```
kevindougherty-noaa: I believe the more "pythonic" way of addressing this conditional would be:

`if levels:`

See here: https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty

src/eva/plotting/emcpy/plot_tools/figure_driver.py
```diff
@@ -65,9 +65,21 @@ def figure_driver(config, data_collections, timing, logger):
...
+                step_vars = levels
+                step_var_name = 'level'
+                title_fill = ' Lv. '
 
             # Set some fake values to ensure the loops are entered
             if variables == []:
```
kevindougherty-noaa: Same comment as above.

kevindougherty-noaa (commented):
Just a small change request in regards to a conditional.

src/eva/plotting/emcpy/plot_tools/figure_driver.py
EdwardSafford-NOAA: Good to know, thanks.  Change made.

src/eva/plotting/emcpy/plot_tools/figure_driver.py
EdwardSafford-NOAA: Change made here and elsewhere in the (previously) unchanged code.

kevindougherty-noaa (approved):
Looks good by me, thanks for making the changes. @CoryMartin-NOAA is out of office until Monday.

EdwardSafford-NOAA:
I note the application and notebook tests are failing because not all required elements (like matplotlib) are not loading on the test platform.  That doesn't seem like it might be something these changes brought about.

EdwardSafford-NOAA:
I found a problem with two of the other test scripts as a result of these changes.  I'm going to shift this PR to draft until I figure that out.

EdwardSafford-NOAA:
Found and fixed the problem.  I had missed a necessary default if neither channels nor levels are in play.  Got lucky that I found the problem so soon.

CoryMartin-NOAA:
I think the test failures now are unrelated (ugh again) but will double check by rerunning an old action

src/eva/data/data_collections.py
```diff
@@ -188,31 +190,47 @@ def get_variable_data_array(self, collection_name, group_name, variable_name, ch
...
+                                      f'Level is not a dimension in Dataset')
+                # Make sure it is a list
+                levels_sel = []
+                levels_sel.append(levels)
+
+                # Create a new DataArray with the requested channels
+                data_array_levels = data_array.sel(Level=levels_sel)
+                return data_array_levels
 
         else:
```
CoryMartin-NOAA: I think this else needs modified now, right? Or at very least the error message that gets written out

src/eva/plotting/emcpy/plot_tools/figure_driver.py
```diff
@@ -65,27 +65,40 @@ def figure_driver(config, data_collections, timing, logger):
...
+            step_var_name = 'channel'
+            title_fill = ' Ch. '
+
+            # Get list of levels, conditionally override step variables
+            levels_str_or_list = batch_conf.get('levels', [])
+            levels = parse_channel_list(levels_str_or_list, logger)
+            if levels:
+                step_vars = levels
+                step_var_name = 'level'
+                title_fill = ' Lv. '
```
CoryMartin-NOAA: personal nit pick, would prefer `' Lev. '` , I think the one extra letter makes it less ambiguous, but curious what others think.

CoryMartin-NOAA (commented):
Thanks for this, this will be a fantastic and really useful addition. I have just a couple comments.

CoryMartin-NOAA:
Regarding #124 I'd have to try it, but theoretically, yes, this should do it.

src/eva/plotting/emcpy/plot_tools/figure_driver.py
EdwardSafford-NOAA: Done.

src/eva/data/data_collections.py
EdwardSafford-NOAA: Fair point.  See if these changes handle things better.

CoryMartin-NOAA (approved):

src/eva/data/data_collections.py
CoryMartin-NOAA: Yep thanks!

CoryMartin-NOAA:
Since this touches the data collections, I would like either @danholdaway or @asewnath to approve before merging in.

asewnath (approved):

danholdaway (approved):

PR #158, opened 2023-09-29
Refactoring for Hvplot batch plotting

asewnath:
## Description

We would like to use eva to create batches of hvplot figures for future web development work. In order to do this, the figure driver needs to be generalized to take in multiple backends. Then, the emcpy and hvplot  backends would be selected based on the `plotting_backend` field in the config yaml. A handler was created to direct each backend to it's associated CreatePlot and CreateFigure classes. 

Please note the new test yaml: `testHvplotIodaObsSpaceAmsuaN19.yaml`. The hvplot backend will expect the same yaml configurations that emcpy does.

Currently, only the scatter plot works for the hvplot backend. We've decided that it would be best to get feedback on the restructuring before we moved on with further hvplot development considering there are a lot of changes. 

The new directory structure has two main directories under plotting: `batch` and `interactive`. Under the `batch`  directory will be a base directory and backend directories that will inherit from the base directory. In the future, all diagnostics will inherit data prepping/processing from the base diagnostic classes.


danholdaway:
I probably wont have time to review before furlough. But just one quick spot and comment. Since the YAMLs do have to change I think we might as well make the backend a required argument. I.e. add EMCpy to all the test YAMLs.

CoryMartin-NOAA:
Ditto for me with what Dan said :-( here's hoping I see GitHub again soon

danholdaway:
Could you update the requirements.txt to include these new packages? I think we should have three requirements packages:
- requirements_github.txt
- requirements.txt
- requirements_emc.txt

The first one used by github stays as is. The second one should be the same as the first one but with >= instead of == and not include the things needed to build the docs. The third _emc would be the same as the current requirements.txt. I.e. not have seaborn, bokeh, hvplot etc. Would that arrangement work for you @CoryMartin-NOAA?

Would it be possible to create a situation where if the chosen backend is hvplot but this package is not available we get a sensible error?

```python
try:
    import hvplot
except ImportError:
    logger.abort("The hvplot backend is not available since hvplot is not in the environment.")
```

CoryMartin-NOAA:
Yes, @danholdaway that sounds perfect.

src/eva/plotting/batch/base/plot_tools/figure_driver.py
```diff
@@ -40,7 +39,28 @@ def figure_driver(config, data_collections, timing, logger):
...
+    # Get plotting backend
+    # --------------------
+    backend = graphics_section.get('plotting_backend')
+
+    if backend == 'Hvplot':
+        try:
+            import hvplot
+        except ImportError:
+            logger.abort("The hvplot backend is not available since hvplot is not in the environment.")
+            backend = 'Emcpy'
```
danholdaway: This line is redundant.

src/eva/plotting/batch/base/plot_tools/figure_driver.py
```diff
@@ -40,7 +39,28 @@ def figure_driver(config, data_collections, timing, logger):
 
     # Get list of graphics from configuration
     # -------------------
-    graphics = config.get("graphics")
+    graphics_section = config.get('graphics')
+    graphics = graphics_section.get('figure_list')
+
+    # Get plotting backend
+    # --------------------
+    backend = graphics_section.get('plotting_backend')
```
danholdaway, line 47: Perhaps there should also be a general check that the chosen backend string matches at least one available backend? Just in case of a typo?

src/eva/plotting/batch/base/plot_tools/figure_driver.py
```diff
@@ -158,12 +180,23 @@ def make_figure(figure_conf, plots, dynamic_options, data_collections, logger):
...
-            layer_list.append(layer_class(layer, logger, data_collections).plotobj)
+
+            # Temporary case to handle different diagnostics
+            if handler.BACKEND_NAME == 'Emcpy':
+                eva_class_name = layer.get("type")
+                eva_module_name = camelcase_to_underscore(eva_class_name)
+                full_module = "eva.plotting.batch.emcpy.diagnostics."+eva_module_name
+                layer_class = getattr(im.import_module(full_module), eva_class_name)
+                layer_list.append(layer_class(layer, logger, data_collections).plotobj)
+            else:
```
danholdaway, line 195: Should it be an elif hvplot?

danholdaway (approved):
Just a couple of very minor comments. Overall this looks great. I like that you have have essentially the same YAML with just one key different and can make the plots using EMCpy and HVplot. Thanks for working on this.

danholdaway:
One other thing. Can you make a minor version bump in setup.py to 1.6.0? Then once we have a fully working hvPLot backend perhaps we make it eva 2.0 since that would be a significant new feature.

danholdaway:
I just pushed a little clean up to the requirements. There were still some lingering things in setup.py that are replaced by what's in all the requirements.txt files. I figure we should minimize the number of places we have these dependencies. I also added a requirements file that is used when installing with spack-stack 1.4

CoryMartin-NOAA (commented):
You've got a swap file in here.

setup.py
```diff
@@ -31,20 +32,6 @@
         'Natural Language :: English',
         'Operating System :: OS Independent'],
     python_requires='>=3.6',
```
CoryMartin-NOAA, line 34: The CI test fails, I wonder if it's not installing properly with these removed?

src/eva/tests/config/testCubedSphereRestart.yaml
```diff
@@ -19,6 +19,9 @@ datasets:
     orography variables: [geolon, geolat]
 
 graphics:
+
+  plotting_backend: Emcpy
+  figure_list:
```
CoryMartin-NOAA, line 24: an extra blank line here to be consistent with the others? (nitpick)

CoryMartin-NOAA (commented):
A few extraneous swapfiles merged in, but beyond that (and figuring out the Actions failure) this looks good to me.

setup.py
danholdaway: I think I forgot to add emcpy to the gh requirements

src/eva/plotting/batch/base/plot_tools/figure_driver.py
asewnath: I can add elif hvplot but technically this if else case wouldn't be in here once every diagnostic is converted to one that is inherited from base. In the else case, the `data_prep()` and `configure_plot()` are invoked and those are methods of the new base classes

asewnath:
> You've got a swap file in here.

image

asewnath:
@danholdaway I just pushed a branch that deleted your changes, and I tried to revert but I don't see files like the spack stack requirements file anymore. Can you revert back to the changes you made and then I can try again with the review changes?

danholdaway:
> @danholdaway I just pushed a branch that deleted your changes, and I tried to revert but I don't see files like the spack stack requirements file anymore. Can you revert back to the changes you made and then I can try again with the review changes?

I think they are still there when I look at https://github.com/JCSDA-internal/eva/tree/feature/bokeh_batch_plots

asewnath:
> > @danholdaway I just pushed a branch that deleted your changes, and I tried to revert but I don't see files like the spack stack requirements file anymore. Can you revert back to the changes you made and then I can try again with the review changes?
> 
> I think they are still there when I look at https://github.com/JCSDA-internal/eva/tree/feature/bokeh_batch_plots

I think something is going on on my end because I'm not seeing your commit in my commit history. Do you mind sharing the commit id?

danholdaway:
Can you 

> > > @danholdaway I just pushed a branch that deleted your changes, and I tried to revert but I don't see files like the spack stack requirements file anymore. Can you revert back to the changes you made and then I can try again with the review changes?
> > 
> > 
> > I think they are still there when I look at https://github.com/JCSDA-internal/eva/tree/feature/bokeh_batch_plots
> 
> I think something is going on on my end because I'm not seeing your commit in my commit history. Do you mind sharing the commit id?

Can you try doing a git pull? Otherwise perhaps just reclone the repo? I'm at commit [d54420c](https://github.com/JCSDA-internal/eva/commit/d54420c03abe784af9270cb7a323c397bcac2364)

danholdaway:
Oh maybe I hadn't pushed everything I had locally. I just did a push and something went.

asewnath:
> Oh maybe I hadn't pushed everything I had locally. I just did a push and something went.

I don't know why, but whatever you did made all of your commits reappear in my commit history (like the one from a few hours ago). Odd github behavior, but we're back on track now

CoryMartin-NOAA (commented):
Looks like the tests pass now with that requirements change, but we still have loitering swapfiles committed.

asewnath:
> Looks like the tests pass now with that requirements change, but we still have loitering swapfiles committed.

Yep, those will be taken care of in just a moment

CoryMartin-NOAA (approved):

Issue #84, opened 2023-03-31
Add support for binary station data files


EdwardSafford-NOAA:
## Description

Extend `src/eva/data/mon_data_space.py` to support binary files in GrADS station format.

Support for legacy Radiance DA monitor files was recently added in `src/eva/data/mon_data_space.py`.  These are simple binary files.  The exact file format (variable names and dimensions) is described by the accompanying control (*.ctl) file.

The legacy Ozone DA monitor (OznMon) data files are in GrADS **station** format.  This is a binary file with a header on each obs record.  The header includes the station data (station id, lat, lon, etc).  The accompanying control file only indicates the variable information and includes the indicator `dtype station`.  The station header does not vary so that format is not specified in the control file.

## Requirements

Given a yaml file identifying an OznMon control file and 4 data files, load the data and generate an image with 4 map-scatter plots each containing the data from one specific cycle.

## Acceptance Criteria (Definition of Done)

A test yaml file correctly loads 4 OznMon data files and generates the 4 pane map-scatter of the data.  The specific variable doesn't matter but I'll use o-f for consistency with the existing OznMon horizontal plots.

## Dependencies
None

Any of those able to do so -- please assign me to this issue.  Thx.

EdwardSafford-NOAA:
After some additional thought I'm not sure this is a high priority.  Maybe we can talk abouit this at the next Mon 2.0 meeting.  I'd like to keep this issue open, but don't expect to work on it in the near term.   It can still be assigned to me though.

EdwardSafford-NOAA:
Update -- I need to take this on now.  I'm ready to add Ozn horizontal plots, and it's likely that some horizontal plots now associated with the ConMon will be desirable as well.

EdwardSafford-NOAA:
I've figured out the structure of the OznMon horiz files.  It's a 28 byte header (station id, lat, lon, time index, nlev, surf flag) followed by 88 floats for the 4 variables and 22 pressure levels.  That is repeated n times and the last header has nlev=0 which signals EOF.  Next step is to map that data to the structure that emcpy needs to make a horizontal plot.

EdwardSafford-NOAA:
One complication is that the OznMon and ConMon control files are significantly different and mon_data_space.py will have to handle both.  Neither actually specifies the number of levels directly.  The OznMon uses this specification:

```
*XDEF is pressure level number
*  x=    1, level=      0.101 , iuse=  1 , error=    0.020
[...]

PR #160, opened 2023-10-16
Add support for DA monitor binary station data files

EdwardSafford-NOAA:
## Description

Extend `data/mon_data_space.py` to support legacy DA monitor binary station data files.  These are a binary files with a header on each observation record. The header includes the station data (station id, lat, lon, etc). The accompanying control file only indicates the variable information but does not include the number of observations.  This can only be learned by reading the data file.

This completes issue #84 .

## Dependencies

None

## Impact

None


src/eva/data/mon_data_space.py
```diff
@@ -77,43 +92,70 @@ def execute(self, dataset_config, data_collections, timing):
...
+
+                # Read station data file.  Note that there is no guarantee the number
+                # of variables is the same for different station data files.  The file_ctr
+                # is used to tack on a number to distinguish the coordinates/dimensions
+                # and corresponding variables for file1, file2, etc.
+                darr, cycle_tm, dims, lat, lon = self.read_stn_ieee(filename, coords, dims,
+                                                                    ndims_used, dims_arr, nvars,
+                                                                    stn_vars)
+                y_range = np.arange(1, dims['ydef']+1)
+                coords['ydef'] = "Nobs" + str(file_ctr)
```
CoryMartin-NOAA: Does this string need a space in it? `"Nobs " + str(file_ctr)`?

src/eva/data/mon_data_space.py
```diff
@@ -77,43 +92,70 @@ def execute(self, dataset_config, data_collections, timing):
...
-                                         dims_arr, x_range, y_range, z_range, cyc_darr, channo)
+                                         dims_arr, x_range, y_range, z_range, channo, cyc_darr)
 
             if attribs['sat']:
                 timestep_ds.attrs['satellite'] = attribs['sat']
             if attribs['sensor']:
                 timestep_ds.attrs['sensor'] = attribs['sensor']
 
+            if len(lat):
+                lat_str = 'lat' + str(file_ctr) if stn_data else 'lat'
```
CoryMartin-NOAA: Similar comments on these lines

CoryMartin-NOAA (commented):
Couple brief questions/comments but otherwise looks good from what I can tell

src/eva/data/mon_data_space.py
EdwardSafford-NOAA: I guess it could, but I don't think it's necessary.  The problem is that the number of obs is file dependent.  So I need a means to associate the variables in a given file with the correct coordinates and dimensions when loading more than one (the ozn horiz plots load 4 files).  This solution gives me variables obs1, ges1, etc associated with dimension nobs1 and coordinate Nobs1, and then a different set with the next file.  Horiz plots typically load 1-4 files.

Alternatively I could figure out the greatest number of obs in the fileset and pad the shorter ones out with NaN values so the dimensions could be shared by all variables. That would be a bit more of more effort in the data load but I'll be glad to go that way if you'd prefer.  Now that I describe it I think I like it better than what I've got now.

CoryMartin-NOAA: Ok I understand better what the string concatenation is doing. I'll defer to you. If you are happy with what you have, so am I but I also will support your suggestion of shared dimensions (it would probably be cleaner)

EdwardSafford-NOAA: Let me give that a try.  If it doesn't take to long it's probably worth it.  I'll switch this PR to draft.

EdwardSafford-NOAA: Revising that was the right answer.  I learned about xarray.align() and the awesome power that wields.  One call syncs the dimensions of all datasets and fills missing values with NaN.  It sure is easy when you know how. :)

CoryMartin-NOAA (approved):
Looks good, thanks!

PR #162, opened 2023-11-09
Extend mon_data_space.py to support MinMon binary data

### LINKED ISSUE #161: Extend mon_data_space.py to support MinMon data
Extend `mon_data_space.py` to support MinMon data.  I wasn't sure this would be necessary but development of a MinMon plotting template for obs-monitor has revealed some things missing from `mon_data_space.py`.  

Reading the MinMon data does not quite work like other sources.  The difference is probably because all monitor data files are generated by Fortran programs, whereas the minmon data files are created by perl using packed, binary format to make them "Fortran-like".  I've figured out a workaround which I need to add.  Also the gnorm plots require the log values of the `allgnorm` variable, but we don't have a log transform available in eva, so I need to add a variable for that.

EdwardSafford-NOAA:
Extend `data/mon_data_space.py` to support MinMon binary data files and 'iteration' as a dimension.  Additionally an error in coordinate assignment was fixed (see 'new_coords' changes in code).

This will close #161 .

### DIFF
--- a/src/eva/data/mon_data_space.py
+++ b/src/eva/data/mon_data_space.py
@@ -9,9 +9,10 @@
 
 # --------------------------------------------------------------------------------------------------
 
-
+import struct
 import os
 import numpy as np
+import array
 
 from xarray import Dataset, concat, merge, align
 from scipy.io import FortranFile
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
looks good to me, thanks!

PR #164, opened 2023-11-17
Add csv_space.py

EdwardSafford-NOAA:
## Description

Add `data/csv_space.py` to load csv data from csv files.  Most immediately this is to be used with the MinMon summary plots, though I've tried to make it as generic as possible to support other uses.  

## Dependencies

None

## Impact

None


src/eva/data/csv_space.py
```diff
@@ -0,0 +1,153 @@
...
+        collection_name = dataset_config['name']
+
+        ds_list = []
+
+        # Read in the CSV files
+        for file in filenames:
+
+            f = open(file, "r")
+            fd = list(csv.reader(f, delimiter=","))
+            file_data = [[element.strip() for element in row] for row in fd]
```
CoryMartin-NOAA, line 46: Can we put in an option that handles if there is a header or not?

src/eva/data/csv_space.py
```diff
@@ -0,0 +1,153 @@
...
+                    dt_arr = self.get_datetime_array(file_data, date_config)
+                    ds = xr.Dataset({var_name: (coord, dt_arr),
+                         coord: range(0, len(dt_arr))})
+                    
+                    ds_list.append(ds)
+
+                # load requested data 
+                if group_vars is not None:
+                    for key, var in group_vars.items():
+                        if coord is None: coord = 'Unit'
```
CoryMartin-NOAA: When will this not be `None`? and why should it always be `'Unit'`?

CoryMartin-NOAA (commented):
Couple of comments to consider this late Friday afternoon, but otherwise this is looking pretty good, thanks @EdwardSafford-NOAA!

src/eva/data/csv_space.py
EdwardSafford-NOAA: It's `None` if there isn't a datetime field (or fields), otherwise it's `Cycle`.  For MinMon purposes it won't ever be `None,` but I was trying to think of how to best to handle things if there wasn't.  I went with `Unit` just because it was generic.  Maybe the coordinate name should be a yaml option?

src/eva/data/csv_space.py
EdwardSafford-NOAA: Yes that's a good idea.  The yaml file will have to specify the header.  I will add that.

src/eva/data/csv_space.py
CoryMartin-NOAA: I like that suggestion

src/eva/data/csv_space.py
EdwardSafford-NOAA: The yaml file can now have a line in the `-groups:` section that can indicate the number of rows in a header:
` header:
          rows: 1
`
These rows are then skipped when the data is converted into the dataset.

src/eva/data/csv_space.py
EdwardSafford-NOAA: The yaml file can now contain an entry like this in the `-groups:` section:

`coordinate:
          name: Fred
`
This coordinate name will be used if there is no datetime information.  If no coordinate is specified and none is provided in the yaml then `Unit` will be used as the default.

CoryMartin-NOAA (approved):
Looks good as far as I can tell, thanks @EdwardSafford-NOAA for another great addition!

kevindougherty-noaa (approved):

PR #179, opened 2024-03-06
Fix levels bug in mon_data_space.py

### LINKED ISSUE #178: Level subset does not work in mon_data_space.py [Bug]
If the `levels `array in the input yaml file does not include all level values found in the control file then the data load fails when the time data is added to the dataset.  The issue is that the several variables associated with levels (`level`, `level_yaxis_z`, `level_assim`, `level_nassim`) need to all be reduced if the actual levels requested is not all available.  This is already done for `Channel` so it should be a simple addition to the `subset_coordinates` method in `mon_data_space.py`.

EdwardSafford-NOAA:
Levels implementation as dataset coordinate in `data/mon_data_space.py` was not entirely complete.  If the input yaml file requested only a subset of the available levels, as opposed to requesting all levels specified in the control file, the results were sub-optimal.  The fix is simple enough -- it now follows the same steps as for the Channel coordinate.

Resolves #178 .

### DIFF
--- a/src/eva/data/mon_data_space.py
+++ b/src/eva/data/mon_data_space.py
@@ -117,6 +117,7 @@ class MonDataSpace(EvaDatasetBase):
                 # will NOT be the same for different station data files.
                 darr, cycle_tm, dims, lat, lon = self.read_stn_ieee(filename, coords, dims,
                                                                     ndims_used, dims_arr, vars)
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
Good old bug fixes, thanks!

PR #181, opened 2024-03-08
Fix level bugs, part 2Ref #180

### LINKED ISSUE #180: mon_data_space.py doesn't handle single levels correctly[Bug]
Level bug redux.  An instrument like ompslp_npp has just one level.  `mon_data_space.py` doesn't correctly handle it.  More than one level is ok, but just one is not working.

EdwardSafford-NOAA:
This PR fixes (fingers crossed) the level bug in `mon_data_space.py`.  These recent bugs all stem from the inconsistent logic in the legacy control files (which describe the structure of the binary station and data files).  Initially I had (naïvely) assumed that if any of the three defined dimensions (xdef, ydef, zdef) were 1 then it could be ignored.  That is not true, since we want to iterate on Level, and that can legitimately be dimension 1 (surface pressure, most ozone time series).  This release corrects that and adds more in-line notes so I'll have a better/faster handle on this if/when I need to revisit it.

Testing has been done with the oznTime plots (from which I encountered the problem, as well as conventional, radiance, and minmon plots to try to test all cases.  

Resolves #180 .

### DIFF
--- a/src/eva/data/mon_data_space.py
+++ b/src/eva/data/mon_data_space.py
@@ -115,9 +115,10 @@ class MonDataSpace(EvaDatasetBase):
 
                 # Read station data file.  Note that the variable dimensions
                 # will NOT be the same for different station data files.
+                # y_range is set here because it's the number of obs which is not in the control
+                # file and is only known once the ieee file is read.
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):
looks good, thanks!

PR #184, opened 2024-03-21
Adding new reader and lat/lon match transform

### LINKED ISSUE #177: Adding capability to compare geovals
We want to add the capability for comparing geovals from different systems (jedi, gsi, geos, etc.) This involves adding a new dataset reader and potentially a transform. The reader would require an obs file along with the geoval file to retrieve lat/lon information. The reader would also take in templated filenames so that it may read more than one instrument file at a time. 

The new transform takes the lat/lon information from `experiment` and `control`, finds a list of indices from `control` that are the closest match to `experiment`, and then updates the `experiment` dataset with variables from the `control` dataset that are index matched to it. The new fields in the `experiment` dataset would look something like this: `experiment_geovals::amsua_n19_from_control_geovals::vegetation_area_fraction`

Potential eva configs for geoval space:

```
datasets:
  - name: experiment_geovals
    type: GeovalSpace
    obs_file:
      - ${data_experiment_path}/{instrument}_experiment.nc4
    geovals_file:
      - ${data_experiment_path}/{instrument}_experiment_geovals.nc4
    levels: &levels 33
    instruments:
      - name: amsua_n19
        geoval_variables: &geoval_variables ['vegetation_area_fraction', 'leaf_area_index']
      - name: avhrr3_metop-b

  - name: control_geovals
    type: GeovalSpace
    obs_file:
      - ${data_control_path}/{instrument}_control.nc4
    geovals_file:
      - ${data_control_path}/{instrument}_control_geovals.nc4
    levels: &levels 33
    instruments:
      - name: amsua_n19
        geoval_variables: &geoval_variables ['vegetation_area_fraction', 'leaf_area_index']
      - name: avhrr3_metop-b

  transforms:
    - transform: index_match
      starting_dataset: control_geovals
      match_index_to_this_dataset: experiment_geovals

```

@CoryMartin-NOAA Please let me know if you have any thoughts or suggestions for this new reader/transform. I had also thought to combine `control` and `experiment` into one dataset read and perform the index matching there so that there's no need for a new transform

asewnath:
We are adding a new reader to handle geoval files, but with the idea that it can be generalized to other types of files. A new latlon match transform was added using some work that @danholdaway had developed.

List of changes:
- Method to `data_collections.py` to retrieve a collection
- Generic `DataFile` reader
- `latlon_match` transform to match lat/lon coordinates from one collection to another
- `DataFile` test yaml and some files for testing

Resolves #177

### DIFF
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -8,7 +8,7 @@ xarray>=2022.6.0
 seaborn>=0.12.2
 hvplot>=0.8.2
 nbconvert>=6.5.4
-bokeh<3.4.0,>=3.2.0
+bokeh<3.5.0,>=3.4.0
 geopandas>=0.13.2
 geoviews>=1.10.0
 nbsite
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -8,7 +8,7 @@ xarray>=2022.6.0
 seaborn>=0.12.2
 hvplot>=0.8.2
 nbconvert>=6.5.4
-bokeh>=3.1.1
+bokeh<3.4.0,>=3.2.0
 geopandas>=0.13.2
 geoviews>=1.10.0
 nbsite
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,8 @@ numpy>=1.22.3
 
 # Not explicitly part of eva but dependcies of eva dependencies already in spack-stack
 # versions need to be set to avoid other versions being picked
-pyproj==3.1.0
+#pyproj==3.1.0
+pyproj
 importlib-metadata==4.8.2
 
 # Additional packages
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,7 @@ numpy>=1.22.3
 
 # Not explicitly part of eva but dependcies of eva dependencies already in spack-stack
 # versions need to be set to avoid other versions being picked
-#pyproj==3.1.0
-pyproj
+pyproj==3.1.0
 importlib-metadata==4.8.2
 
 # Additional packages
--- a/src/eva/data/data_collections.py
+++ b/src/eva/data/data_collections.py
@@ -169,6 +169,11 @@ class DataCollections:
 
     # ----------------------------------------------------------------------------------------------
 
+    def get_data_collection(self, collection_name):
+        return self._collections[collection_name]
+
+    # ----------------------------------------------------------------------------------------------
+
     def get_variable_data_array(self, collection_name, group_name, variable_name,
                                 channels=None, levels=None, datatypes=None):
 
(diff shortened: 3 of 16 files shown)


danholdaway:
Thanks for adding more capability to Eva @asewnath. I'm looking at the test files that you added and to me they look a lot like the files we already read with other classes, either `ioda_obs_space` or `gsi_obs_space`. Rather than adding a new class for reading them can we make use of those classes instead or was there something missing from them? If we do need a new class I would advocate for more descriptive name so indicate the scope of what it can read.

asewnath:
Hi @danholdaway, the test file using the new reader is the geoval file. Since this geoval file has a different format compared to ioda and gsi files (along with level dimensions to consider), a new reader was created for it. In the issue mentioned in this PR, @CoryMartin-NOAA suggested that I generalize the new reader’s name. This 
is to use the reader for other types of similarly structured files later on.

The obs file test file I have added is read by an existing reader. This is primarily to extract lat/lon fields to be used by the match transform later on.

danholdaway:
How straightforward (or not) would it be to allow the generalization of the gsi_obs_space reader to allow to levels in the variables? In my mind we've designed Eva to have specific readers with generic transforms and plotting; the reason being that plotting and transforms are naturally quite generic whereas readers tend to be naturally quite specific and dependent on the files. However there are a limited number of formats that really need to be read. Trying to write generic readers will (I suspect) result in code with lots of options and different YAML keys that will make them confusing and difficult to document. Further it will mean that you have to pass lots of arguments and so they wont lend themselves well to the APIs used in interactive mode. I think you should be able to pass a filename only and be able to read a file. Rather than pass a filename as well as information about coordinate names etc. Does that make sense or am I missing the point here?

 So I would lean towards generalizing the gsi reader which is already geared towards this file format. If we really want to add another reader to handle gsi_obs_space files that have GeoVaLs that's fine with me but I would still advocate for giving it a specific name and keeping it's scope quite narrow with minimal YAML so it's in keeping with the design of the readers in general.

asewnath:
Okay, I can look into generalizing the gsi_obs_space reader

CoryMartin-NOAA:
@danholdaway @asewnath I think the gsi_obs_reader will be quite different from this geoval reader. Mainly because of the code put in there to do channel subsetting for the radiance observations. I do agree though that if it is a lot of work to generalize then it is probably not worth it.

asewnath:
@danholdaway @CoryMartin-NOAA Then maybe we can just make the new reader name more geoval specific for now?

CoryMartin-NOAA:
I'm fine with that

danholdaway:
OK, that works.

danholdaway (approved):

PR #186, opened 2024-05-13
Adding the hvplot backend

asewnath:
## Description

This PR expands upon the hvplot backend introduced in #158. All other diagnostic types are reformatted to have a base class and child classes for each backend. Each base class includes two methods: `data_prep` and `configure_plot`. `configure_plot` is a virtual method that is resolved once a child class for either emcpy or hvplot is instantiated.

`figure_driver` is updated to reflect these changes


asewnath:
Moving github action fix to a different PR since it's still not resolved

CoryMartin-NOAA (approved):
Very cool, thanks!

src/eva/plotting/batch/emcpy/diagnostics/emcpy_density.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class EmcpyDensity(Density):
+
+    """
+    EmcpyDensity class inherits from the Density class and provides methods
+    to configure plotting settings for density plots using the emcpy library.
+
+    Attributes:
+        plotobj: A declarative plotting density object for emcpy density plots.
```
EdwardSafford-NOAA: Do the attributes in the Emcpy* class declarations need to be updated?  The equivalent hvplot class declarations describe the parent class in their Attributes sections.

EdwardSafford-NOAA (approved):
Changes look good, just had the one question on the documentation in the Emcpy* classes.

src/eva/plotting/batch/emcpy/diagnostics/emcpy_density.py
asewnath: Yes, I can update the emcpy documentation to resemble the hvplot declarations. Thanks!

danholdaway (approved):
Thanks Akira. It's great to have a complete backend utilizing hvplot. For future reference you should use `git mv` to rename files so their history is preserved. I'm concerned that the codes aren't tested and so the interfaces could get broken. Do you plan to add some testing through the notebook style of testing?

asewnath:
> Thanks Akira. It's great to have a complete backend utilizing hvplot. For future reference you should use `git mv` to rename files so their history is preserved. I'm concerned that the codes aren't tested and so the interfaces could get broken. Do you plan to add some testing through the notebook style of testing?

Hi Dan, thanks for the comment. I'll add hvplot testing in another PR

PR #187, opened 2024-05-14
Add Time Series Capability

### LINKED ISSUE #166: Time series in EVA?
I would like to start a discussion regarding `EVA` handling timeseries and creating plots with time axes. Creating time series plots is something I started doing with JEDI outputs on our end but I would like to do it in a generic way if there is demand/need. I had some brief discussions here with @danholdaway and @asewnath (she suggested that I create an issue) and I would really appreciate EMC's input on this so I'm tagging all who might be interested: @CoryMartin-NOAA @EdwardSafford-NOAA @kevindougherty-noaa @guillaumevernieres @ADCollard.

A simple example plot is mean (obs - bkgr) over multiple time steps in observation space. `IodaObsSpace` class is already handling JEDI outputs (location, channel dimensions) in observation space so this would be only a matter of adding a `dateTime` dimension. The issue with EVA is that there are no time series handling and an improvement to `EVA` in this regard would benefit everyone in terms of DA monitoring. `EVA` currently reads data, makes necessary transforms, and makes the plots inside a certain folder, say `geos_ocean` for our case, at a single cycle.

I have issues in terms of cycling and file storage so I have to erase JEDI outputs frequently for high resolution simulations that spans couple of months. Hence, ideally `EVA` needs to handle files during the active cycle before they get erased. This may or may not be relevant for the other developers.

### Below is what I'm suggesting (based on our workflow):

For Swell workflow, our `run` directory currently looks like this (for ocean-only DA cycling) :

```
├── run
│   ├── 20210601T120000Z
│   │   └── geos_ocean
│   ├── 20210601T180000Z
│   │   └── geos_ocean
│   ├── 20210602T000000Z
│   │   └── forecast
|   |   └── geos_ocean
```

`forecast` directory contains GEOS related files/outputs whereas `geos_ocean` has JEDI related configs/outputs. `EVA` runs inside `geos_ocean`  at each cycle and produces _fabulous_ plots.

My suggestion is having an extra folder (call it diagnostics or holding) on the same level as time:

```
├── run
│   ├── diagnostics
│   ├── 20210601T120000Z
│   │   └── geos_ocean
│   ├── 20210601T180000Z
│   │   └── geos_ocean
│   ├── 20210602T000000Z
│   │   └── forecast
|   |   └── geos_ocean
```

So `EVA` would process `IODA` outputs and create netcdf file(s) within the `diagnostics` folder with time (and channels etc. if needed) dimension(s). Afterwards, it may append and update these file, or create a new one every so often, you get the idea. This would require some capabilities within EVA in addition to datasets, transforms, graphics, such as write to output (which may be a time sink unless handled in parallel). It would be great addition with the `EVA` interactive tool as part of DA monitoring.

I'm open to suggestions, thoughts, criticisms..

danholdaway:
## Description

Add capability to read and transform on the fly and assemble dataset that has a time series dimension.

- Add test for compiuting mean of omb and making time series.
- Add test for plotting bias correction coefficients time series.
- Add var bc data.
- Add reader for JEDI bias and tlapse files.
- Removed prints of datasets from read and transform classes and added to driver level.

Plots the new tests create shown below. Note that the tests are somewhat contrived since the data is the same at both time steps. We can try to improve on that in the future.

time_series_omb

varbc_time_series

## Dependencies

NONE

## Impact

NONE


CoryMartin-NOAA:
@danholdaway did the CI test fail because of yet another python version/dependency issue in the container?

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
+# (C) Copyright 2021-2022 NOAA/NWS/EMC
```
CoryMartin-NOAA: 2024

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
...
+import xarray as xr
+
+from eva.data.eva_dataset_base import EvaDatasetBase
+from eva.utilities.config import get
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+class JediVariationalBiasCorrection(EvaDatasetBase):
```
CoryMartin-NOAA, line 23: FYI I think this will change very soon with the changes for aircraft VarBC

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
...
+        for key in required_keys:
+            self.logger.assert_abort(key in dataset_config, "For JediVariationalBiasCorrection " +
+                                     f"the config must contain key: {key}")
+
+        # Parse config
+        collection_name = dataset_config['name']
+        bias_file = get(dataset_config, self.logger, 'bias_file')
+        lapse_file = get(dataset_config, self.logger, 'lapse_file')
+
+        print(bias_file)
```
CoryMartin-NOAA: Do we want this formatted more elegantly or removed completely?

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
...
+        predictor_names = bias_dataset['predictors'].values
+
+        # Initialize a dictionary to store the new data variables
+        new_data_vars = {}
+
+        # To eva-rise the dateset we dont want predictor as a dimension, otherwise all the plotting
+        # code would have to be made to understand that as a dimension and make that part less
+        # generic. Instead we want more variables with predictor as the variable name and the type
+        # as the group.
+        vars_to_flatter = [
```
CoryMartin-NOAA: these are changing to be camelCase

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
...
+
+        print(bias_file)
+
+        # Read the contents of the bias file (netCDF) into an xarray dataset
+        bias_dataset = xr.open_dataset(bias_file)
+
+        # Assign proper coordinates to the dataset and remove nchannels and npredictors from
+        # variables
+        bias_dataset = bias_dataset.assign_coords(
+            {'nchannels': range(len(bias_dataset.nchannels)),
```
CoryMartin-NOAA: These are changing

src/eva/data/jedi_variational_bias_correction.py
```diff
@@ -0,0 +1,168 @@
...
+
+        Example:
+            ::
+                    ioda_instance = JediVariationalBiasCorrection()
+                    default_config = ioda_instance.generate_default_config([bias_file, lapse_file],
+                                                                           collection_name)
+        """
+
+        return {
+            'bias_file': filenames[0],
```
CoryMartin-NOAA, line 151: See: https://github.com/orgs/JCSDA-internal/discussions/118. My main comment is ensure that this is not already out of date

src/eva/eva_driver.py
```diff
@@ -9,16 +9,170 @@
...
+        'interval',
+        'collection',
+        'variables',
+        ]
+    for key in required_keys:
+        logger.assert_abort(key in eva_dict['time_series'], 'If running eva in time series ' +
+                            f'mode the time series config must contain "{key}"')
+
+    # Write message that this is a time series run
+    logger.info('This instance of Eva is being used to accumulate a time series.')
```
CoryMartin-NOAA, line 111: random, but above line is `eva`, this line is `Eva`, we should be consistent. Let's call it "technical debt", and assign @guillaumevernieres to fix it all for us :-)

src/eva/eva_driver.py
```diff
@@ -9,16 +9,170 @@
...
+    # Make list of dates from begin to end with interval
+    dates = []
+    date = begin_date
+    count = 0
+    while date <= final_date:
+        dates.append(date)
+        date += interval
+        count += 1
+        # Abort if count hits one million
+        logger.assert_abort(count < 1000000, 'You are planning to read more than one million ' +
```
CoryMartin-NOAA:
```suggestion
        logger.assert_abort(count < 1e6, 'You are planning to read more than one million ' +
```

src/eva/time_series/__init__.py
```diff
@@ -0,0 +1,9 @@
+# (C) Copyright 2021-2022 United States Government as represented by the Administrator of the
```
CoryMartin-NOAA: NASA?!? :-)

src/eva/data/jedi_variational_bias_correction.py
danholdaway: Thanks for flagging that. We should definitely be in the now in terms of format. I've updated the PR.

danholdaway: Note, though, that for now I'm squeezing the record dimension as that might require more thought for Eva. Essentially Eva want variables to have at most two dimensions and where there are two dimensions one should be Channel. If we have Record in the future we might need to think a bit.

src/eva/tests/config/testIodaObsSpaceAmsuaN19_TimeSeries.yaml
```diff
@@ -0,0 +1,73 @@
...
+        variables: &variables [brightnessTemperature]
+      - name: hofx
+  - name: experiment
+    type: IodaObsSpace
+    filenames:
+      - ${data_input_path}/ioda_obs_space.amsua_n19.hofx.2020-12-14T210000Z.nc4
+    channels: *channels
+    groups:
+      - name: ObsValue
+      - name: hofx
```
Dooruk, line 20: Is there a reason these two experiment groups are repeating? Is one supposed to be 03z on the 15th?

src/eva/utilities/duration.py
```diff
@@ -0,0 +1,63 @@
...
+        r'(?:(?P<years>\d+)Y)?'  # years
+        r'(?:(?P<months>\d+)M)?'  # months
+        r'(?:(?P<days>\d+)D)?'  # days
+        r'(?:T'  # starts time part with 'T'
+        r'(?:(?P<hours>\d+)H)?'  # hours
+        r'(?:(?P<minutes>\d+)M)?'  # minutes
+        r'(?:(?P<seconds>\d+)S)?'  # seconds
+        r')?'  # end time part
+    )
+
```
Dooruk, line 40: Huh, I didn't realize `isodate` wasn't part of the JEDI spack.

src/eva/time_series/time_series.py
```diff
@@ -0,0 +1,100 @@
...
+
+import numpy as np
+import xarray as xr
+
+
+# --------------------------------------------------------------------------------------------------
+
+
+xr_aggregation_methods = {
+    'mean': lambda ds, dim: ds.mean(dim=dim, skipna=True),
```
Dooruk, line 21: I'm curious if there would be huge performance differences between using lambda vs. ufuncs, especially while dealing with large IODA outputs. When I processed SSS IODA output (~350k obs) mean operation was taking long.

Dooruk (approved):
Thanks @danholdaway for adding this feature. I'm looking forward to test this in the workflow level.

src/eva/tests/config/testIodaObsSpaceAmsuaN19_TimeSeries.yaml
danholdaway: Just to avoid adding more data to the repo for testing.

src/eva/utilities/duration.py
danholdaway: It might well be but it's not a standard library so I would prefer not to depend on it for something so simple. Not such an issue for GMAO but we might otherwise have to get the software approved for the operational machine and it's not worth it.

src/eva/time_series/time_series.py
danholdaway: In this case I would not expect a performance bottleneck from the lambda. While it might be costly if embedded in a larger loop I think here it would be OK and dwarfed by the cost of the actual aggregation method. What kind of cost are you seeing for your case? Could you try some vanilla Python/Xarray that computes the aggregation you want and let me know if the cost is much different than running with Eva?

asewnath (approved):
I was able to create a test plot for our 3dvar_atmos run, thanks for adding this feature into eva!

src/eva/time_series/time_series.py
Dooruk: Sorry, I meant when I processed with an offline xarray code that I wrote before. 

Let us implement this in Swell and we can see performance with the full obs system. I don't want to hold back this merge longer since you will be on leave for a while.

danholdaway: Ok sounds good.

PR #199, opened 2024-07-05
Fix time precision warning on hera

### LINKED ISSUE #190: Data type conversion warning in csv_space.py
## Description

The updated python version on hera now reports this warning in `csv_space.py`:

`0: /scratch1/NCEPDEV/da/python/envs/eva/lib/python3.10/site-packages/eva/data/csv_space.py:70: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
`

I'll take a look at this when I've got a moment.

## Dependencies

None

EdwardSafford-NOAA:
With the rocky8 update `csv_space.py` produced a warning message about the precision of datatime64 data.  Using nanosecond precision formatting fixes this warning.

Testing has been done on both hera and wcoss2 (which didn't have any issue with the datetime64 format and is ok with nanosecond formatting).

Closes #190

### DIFF
--- a/src/eva/data/csv_space.py
+++ b/src/eva/data/csv_space.py
@@ -156,4 +156,6 @@ class CsvSpace(EvaDatasetBase):
                               " \'year\': int, \'month\': int, \'day\': int, and \'hour\': int. " +
                               f" Date information found was {date_config}")
 
-        return np.array([np.datetime64(datetime.strptime(ds, '%Y%m%d%H')) for ds in date_str_list])
+        return np.array([np.datetime64(datetime.strptime(ds,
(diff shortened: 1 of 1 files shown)


EdwardSafford-NOAA:
@CoryMartin-NOAA and this one too is good to go.

CoryMartin-NOAA (approved):

PR #204, opened 2024-07-30
Fix config error with emcpy_horizontal_line.py Ref

### LINKED ISSUE #203: [Bug] emcpy_horizontal_line.py uses the wrong config
In `emcpy_horizontal_line.py` (line 46) the plot object's config information is modified to remove things that emcpy won't understand, but then the original information (self.config) is returned to the plot object rather than using the `new_config`.  This results in a problem in emcpy because `type` is not a recognized variable and an exception is thrown.

From what I can tell this is a fairly long-standing bug.  I'm surprised I haven't run into it heretofore.  It was an adventure to find but a simple fix.

@kevindougherty-noaa this is the source of that problem I've been having since last week that I thought might have been a build issue.

EdwardSafford-NOAA:
Fix horizontal line config error.  Use the correct config information in the plot object.

I ran into this issue last week but assumed it had to be an install issue, not a simple coding error.  I'm not sure how/why this worked as long as it did.  Maybe the unrecognized variable in the plot object is ok in python3 environments that aren't on Rocky8? 

Close #203

### DIFF
--- a/src/eva/plotting/batch/emcpy/diagnostics/emcpy_horizontal_line.py
+++ b/src/eva/plotting/batch/emcpy/diagnostics/emcpy_horizontal_line.py
@@ -43,7 +43,8 @@ class EmcpyHorizontalLine(HorizontalLine):
         delvars = ['type', 'schema']
         for d in delvars:
             new_config.pop(d, None)
-        self.plotobj = update_object(self.plotobj, self.config, self.logger)
+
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA (approved):

CoryMartin-NOAA:
Gotta love those one line bugs!

PR #209, opened 2024-10-08
Multiple Timeseries

asewnath:
## Description

Extending time series utility to process multiple time series. This allows for time series comparisons between different collections Note that time series section must be a list.


CoryMartin-NOAA (approved):
Looks good, thanks for adding this!

PR #211, opened 2024-11-12
Allow missing/empty cycle time files in timeseries

asewnath:
## Description

- Added support for missing/empty cycle time files when running in timeseries mode
- Adding requirements that work on discover milan nodes

## Demonstration
Comparing two time series of different lengths:
time_series_omb

Timeseries with a missing cycle time:
time_series_omb


requirements.txt
```diff
@@ -1,18 +1,19 @@
 setuptools>=59.4.0
 pyyaml>=6.0
 pycodestyle>=2.8.0
-netCDF4>=1.5.3
-matplotlib>=3.9.0
+netCDF4
```
danholdaway, line 4: I'm not sure I like the idea of pinning versions of libraries in a generic requirements.txt. It could break the installs of Eva on different platforms.

danholdaway (changes requested):
Could you add some testing to show the use case of this feature?

asewnath:
> Could you add some testing to show the use case of this feature?

Yes, this PR is currently a draft

danholdaway (approved):
Nice feature and thanks for adding testing.

PR #213, opened 2024-11-19
Fix plot logo and tight layout bug

### LINKED ISSUE #212: [Bug] Batch plots fail when trying to use 'plot logo' and 'tight layout'
## Current behavior (describe the bug)

Batch plots which include either `plot logo` or `tight layout` in the input yaml file fail catastrophically with a report that matplotlib doesn't understand what it's been handed.  

The issue is that `eva/src/eva/plotting/batch/base/plot_tools/figure_driver.py` correctly calls the `fig.plot_logo()` and `fig.tight_layout()` methods but the corresponding dictionary entries are left in `figure_conf`.  This eventually leads to the confusion in matplotlib and the fatal error.

## To Reproduce

What computer are you running on?  
Problem is not platform specific; it fails equally well on wcoss2 and hera.

Steps to reproduce the behavior

In eva/src/eva/testing/config run this command:
> eva testMonDataSpaceHirs4Metop-A.yaml

EdwardSafford-NOAA:
Fix the fatal error that results from a batch plot input yaml file containing `plot logo` and/or `tight layout` figure settings.

As described in #212 the problem is that in `src/eva/plotting/batch/base/plot_tools/figure_driver.py` both `plot logo` and `tight layout` are left in the `figure_conf` dictionary after the corresponding figure method is correctly applied.  The `figure_conf` dictionary is eventually fed to `emcpy` and then `matplotlib` which is unable to parse these entries.  The fix is to simply remove the entries from `figure_conf` once the requested method has been called.

Testing has been done on hera using `src/eva/tests/config/testMonDataSpaceHirs4Metop-A.yaml`, which now can plot logos and use tight layout.

Closes #212

### DIFF
--- a/src/eva/plotting/batch/base/plot_tools/figure_driver.py
+++ b/src/eva/plotting/batch/base/plot_tools/figure_driver.py
@@ -241,8 +241,11 @@ def make_figure(handler, figure_conf, plots, dynamic_options, data_collections, 
             fig.tight_layout(**figure_conf['tight layout'])
         else:
             fig.tight_layout()
+        figure_conf.pop('tight layout')
+
     if 'plot logo' in figure_conf:
         fig.plot_logo(**figure_conf['plot logo'])
+        figure_conf.pop('plot logo')
 
     saveargs = get_saveargs(figure_conf)
     fig.save_figure(output_file, **saveargs)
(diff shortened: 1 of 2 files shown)


CoryMartin-NOAA (approved):
Looks good, I want @kevindougherty-noaa to review before merging just to make sure it looks okay

kevindougherty-noaa (approved):
Looks good to me as well.

PR #220, opened 2025-03-17
Add files needed for contour filled map plots and gridded plots

kevindougherty-noaa:
## Description

This PR looks to add a missing need for Filled Contour Map plots and Gridded plots. Specifically attempting to solve the following issue: [here](https://github.com/JCSDA-internal/eva/pull/216#issuecomment-2672667506)

## Dependencies

Waiting on the following PRs:
- [x] waiting on [NOAA-EMC/emcpy/pull/145](https://github.com/NOAA-EMC/emcpy/pull/145)


kevindougherty-noaa:
The updated scripts add map filled contour plotting capabilities. The updated test yaml in this PR creates the following figures:
image
image

@Dooruk is this what you are looking for?

Dooruk:
@kevindougherty-noaa thank you for looking into this. I'm looking to plot a vertical transect.

For instance:

longitude(x)
depth(z)
temp(x,z)

pcolormesh(longitude, depth, temp, **kwargs)

**Note**: Getting to these dimensions would require an additional EVA transform, something like `isel` or `sel` from xarray but let's ignore that for now.

Which would then look something like the following:

noDA_transect

I'm not sure which plot type from emcpy I should be using for this, that's what I was trying with filledContour

kevindougherty-noaa:
@Dooruk Thank you for your patience as I was on leave last week. What you are looking for is EMCPy's `GriddedPlot()` function. This uses `pcolormesh()` in the backend to plot, similar to what you are using above. If you share the data you are using, I would be happy to try and replicate this figure using `eva`.

Dooruk:
@kevindougherty-noaa, do you need me to provide you a netcdf file?

I don't have the data I provided in EVA context yet.

kevindougherty-noaa:
@Dooruk I spoke with Cory and I will provide you a solution similar to what you are looking using an FV3 file I have available. Working on it now.

Dooruk:
Thanks, I'm working on a select_dimension transformation implementation in the mean time

kevindougherty-noaa:
@Dooruk I have made the appropriate adjustments and additions to `eva` plotting from `emcpy`. I know you have added your transform PR, but I believe we can accomplish what you are looking for with a slice of the data. It took some time due to a silly oversight on my end with a missing data assignment which took a few days to debug. Here is an example adjusting the `testLatLon.yaml`. It is only looking at a few levels because that is all the data has, but it is the same concept as your figure above. The only caveat with slicing is you need to know the index of the data where you want to look. For example [:,48:,:] is looking at the closest latitude to the equator as the data does not have a true 0 latitude:

```
datasets:
  - group: increment
    type: LatLon
    filename: /scratch1/NCEPDEV/da/Kevin.Dougherty/eva/src/eva/tests/data/gdas.t00z.atminc.nc
    name: experiment
    variables: [T_inc, lev, lon]

graphics:

  plotting_backend: Emcpy
  figure_list:

  # Map plots
  # ---------

  # Observations
  - batch figure:
      variables: [T_inc]
    figure:
      figure size: [20,10]
      layout: [1,1]
      title: 'Vertical Transect of Temperature Increment from GSI'
      output name: gridded_plots/increments/${variable}/vert_transect_inc_${variable}.png
    plots:
      - add_colorbar:
          label: Temperature
          orientation: vertical
        add_xlabel: Longitude
        add_ylabel: Level
        layers:
        - type: GriddedPlot
          x:
            variable: experiment::increment::lon
          y:
            variable: experiment::increment::lev
          z:
            variable: experiment::increment::T_inc
            slices: '[:,48,:]'
          label: T increment
          colorbar: true
          cmap: 'coolwarm'
          vmin: -1
          vmax: 1
```

which produces the following figure:

image

CoryMartin-NOAA:
@kevindougherty-noaa can you add your example to the test YAML so that it lives in the repository and is tested going forward?

Dooruk (approved):
Thanks! I haven't tested it but looks like what I asked for. My PR adds some changes in terms of selecting lat/lon and adding a depth dimension.

Sorry for multiple GMAO requirement files, I will get rid of some we don't use/need.

CoryMartin-NOAA (approved):

PR #228, opened 2025-04-17
updates to optimize current scripts

kevindougherty-noaa:
## Description

Updates to scripts to improve efficiency.

Updates in `figure_driver.py` and the `make_figure()` function avoids some repeated logic, adds some fail safes to make the code more robust and makes the figure structure a little neater. Some steps have helped with optimization of code as well. I have noted in the code that there are areas where parallelization can be utilized given an option in the configure YAML files, but did not add the infrastructure yet. If this is something that may be of interest, especially if we move forward with this work to replace all monitoring efforts, it could be a good option if we are plotting hundreds of plots. 

## Dependencies

If there are PRs that need to be merged before or along with this one, please list them below
Waiting on the following PRs:
- [x] waiting on [emcpy #147](https://github.com/NOAA-EMC/emcpy/pull/147)

## Impact

N/A

### DIFF
--- a/src/eva/data/lat_lon.py
+++ b/src/eva/data/lat_lon.py
@@ -38,11 +38,7 @@ class LatLon(EvaDatasetBase):
                               f' The valid types are {valid_groups}')
 
         # open the input netCDF file
-        ds = xr.open_dataset(filename)
-
-        # Drop data variables not in user requested variables
-        vars_to_remove = list(set(list(ds.keys())) - set(variables))
-        ds = ds.drop_vars(vars_to_remove)
+        ds = xr.open_dataset(filename, engine='netcdf4')[variables]
 
         # rename variables in dataset_config
         rename_dict = {}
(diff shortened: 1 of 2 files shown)


karpob:
It would be awesome to get multiprocessing for batch figures. This is really helpful for doing timeseries plots of hyperspectral sensors where you may have one plot/channel.

kevindougherty-noaa:
@karpob I agree! I discussed with @CoryMartin-NOAA offline about this and we plan to add this feature in a future PR.

karpob:
@kevindougherty-noaa excellent! Possibly a separate issue, but it would be nice to add the ability to incorporate dask to maybe help with parallelization of crunching numbers. Some of the manipulation in  the xarray/data structures may be problematic (last I looked times series looked most problematic, possibly). The way our current legacy "operational" radiance monitoring/timeseries works is making a pile of text files in parallel, then makes plots using multiprocessing. It makes it relatively fast, but super messy. Having eva be a bit slower makes it a little bit of a "hard sell" to move people away from the legacy system.

Dooruk:
> It would be awesome to get multiprocessing for batch figures. This is really helpful for doing timeseries plots of hyperspectral sensors where you may have one plot/channel.

Sorry, I will seem nosey but I saw this because I'm signed up to the all EVA notifications.

Dan and Akira utilized `Pool` for multiprocessing for EVA on a **single node** for Swell, it works both for head and compute nodes:

https://github.com/GEOS-ESM/swell/blob/ffd28ad14282c1c4a6f9c2d1f992ddfdc59e217f/src/swell/tasks/eva_observations.py#L11

Given Milan nodes are 128 cores, this approach may be useful for your workflow for the short term?

CoryMartin-NOAA (approved):

PR #230, opened 2025-04-22
Modify jedi_variational_bias_correction to support multiple data files in input yaml specification

### LINKED ISSUE #225: Modify JediVariationalBiasCorrection to support multiple files within a specified dataset
## Description

The input yaml file contains one or more dataset definitions.  Most `eva` data objects can process multiple filenames within a single dataset definition, but `JediVariationalBiasCorrection` cannot.  Multiple files for use in the same experiment::group::variable must be staged in separate dataset specifications with a single filename for `bias_file` and `lapse_file`.  This works but it becomes cumbersome if more than 2 time periods are used in the plot.  Instead, the `JediVariationalBiasCorrection` object should be modified to treat the input filenames as lists and iterate over them to load the requested data.

## Dependencies

None.

EdwardSafford-NOAA:
Allow jedi_variational_bias_correction object to support multiple input files per group in the input yaml file.  The bias and lapse files are now loaded into an intermediate dataset group structure iteratively before the dataset group is added to the main data_collection object.  This follows similar patterns in the various data objects that support multiple input files within a specified experiment/group.

Additionally catch a mismatch in the number of specified bias and lapse files and abort with appropriate error message.

Closes #225

### DIFF
--- a/src/eva/data/jedi_variational_bias_correction.py
+++ b/src/eva/data/jedi_variational_bias_correction.py
@@ -62,8 +62,8 @@ class JediVariationalBiasCorrection(EvaDatasetBase):
         # Check for required keys in config
         required_keys = [
             'name',
-            'bias_file',
-            'lapse_file',
+            'bias_files',
+            'lapse_files',
             ]
         for key in required_keys:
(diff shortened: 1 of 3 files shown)


CoryMartin-NOAA (approved):
Looks good thanks!

PR #238, opened 2025-05-02
Fix bug in transform collection name sourcing

### LINKED ISSUE #237: [Bug] Time series fails when the channel_stats transform is used
## Current behavior (describe the bug)

If an input yaml file contains both a `time_series:` declaration and a `channel_stat` transform eva fails with this error:
```
 File "/scratch1/NCEPDEV/da/Edward.Safford/noscrub/eva/opt/eva/eva_driver.py", line 164, in read_transform_time_series
    name = transform['new name'].split('::')[0]
KeyError: 'new name'
~~~

What's going on is that after a `time_series` is identified in the input yaml file, the file is scanned for any transforms that are assigned to the same collection as the `time_series`.   Most transforms assign the results to a new variable identified in the transform as 'new name'.  But `channel_stats` don't use 'new name'.  They identify the variable to be processed as 'variable_name'.  So `eva_driver.py` needs to check for that before declaring a fatal error.  Easy fix.

EdwardSafford-NOAA:
Fix bug in check on transform collection name.  The search was done using 'new name' which is present in _most_ transforms but it isn't in `channel_stats`.  Add a check that picks up the collection name for `channel_stats` transforms.

Closes #237

### DIFF
--- a/src/eva/eva_driver.py
+++ b/src/eva/eva_driver.py
@@ -159,8 +159,14 @@ def read_transform_time_series(logger, timing, eva_dict, data_collections):
         transform_dict = defaultdict(list)
         if 'transforms' in eva_dict:
             for transform in get(eva_dict, logger, 'transforms'):
+
                 # Get collection name
-                name = transform['new name'].split('::')[0]
+                name = None
+                if 'new name' in transform:
+                    name = transform['new name'].split('::')[0]
+                elif 'variable_name' in transform:
+                    name = transform['variable_name'].split('::')[0]
+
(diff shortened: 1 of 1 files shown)


CoryMartin-NOAA:
UGH, looks like yet another irrelevant CI issue that needs sorted out

CoryMartin-NOAA (approved):

CoryMartin-NOAA:
I'm just going to force this through. We need a new testing solution, as this is not sustainable to constantly chase these bugs CCing @danholdaway for awareness

PR #236, opened 2025-04-28
Add IODA stats reading capabilities

kevindougherty-noaa:
## Description

Adds capability to read IODA stats files into eva. It is similar to IodaObsSpace, however it has a few different tweaks to handle the data format. This PR also includes a few small changes to `data_collections.py` as well as some plotting scripts.

## Dependencies

None.

## Impact

N/A

PR #244, opened 2025-05-27
Generate filenames from a template

kevindougherty-noaa:
## Description

Within a given eva yaml file, this update allows a list of yaml files to be created given a template, start/end date, interval time, and datetime format. The purpose of this is to process multiple files without having to manually add all the filenames in the yaml files.

## Dependencies

If there are PRs that need to be merged before or along with this one, please list them below
Waiting on the following PRs:
- [ ] waiting on [eva #236](https://github.com/JCSDA-internal/eva/pull/236)

## Impact

None.


kevindougherty-noaa:
Once eva #236 is merged, I will merge develop and all of the extra code should go away.

src/eva/tests/config/testMonSummary.yaml
```diff
@@ -12,12 +12,12 @@ datasets:
...
     control_file:
       - ${data_input_path}/time.hirs4_metop-a.ctl
-    filenames:
-      - ${data_input_path}/time.hirs4_metop-a.2015051418.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051500.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051506.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051512.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051518.ieee_d
+    filenames_template:
+        template: ${data_input_path}/time.hirs4_metop-a.{datetime}.ieee_d
```
CoryMartin-NOAA: What if I want to do something like this:
```
template: gdas.%Y%m%d/%H/atmos/gdas.t%Hz.atmanl.nc
```
Would this work or does it only insert it in one spot in the filename?

kevindougherty-noaa: That is a good point. I updated the code to not take a format and rather just rely on the formatting in the template. Easier and less ambiguous for the user.

src/eva/tests/config/testMonSummary.yaml
```diff
@@ -12,12 +12,11 @@ datasets:
...
       - ${data_input_path}/time.hirs4_metop-a.ctl
-    filenames:
-      - ${data_input_path}/time.hirs4_metop-a.2015051418.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051500.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051506.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051512.ieee_d
-      - ${data_input_path}/time.hirs4_metop-a.2015051518.ieee_d
+    filenames_template:
+        template: ${data_input_path}/time.hirs4_metop-a.%Y%m%d%H.ieee_d
+        start: "2015051418"
```
CoryMartin-NOAA: What are people's thoughts on using the ISO time stamp format instead for these?
```
start: 2015-05-14T18:00:00Z
```

src/eva/utilities/utils.py
```diff
@@ -474,3 +475,93 @@ def is_number(s):
...
+            "template": "gdas.%Y%m%d/%H/atmos/gdas.t%Hz.atmanl.nc",
+            "start": "2024010100",
+            "end": "2024010112",
+            "interval_hours": 6,
+            "datetime_format": "%Y%m%d%H"
+        }
+
+        ➜ ["gdas.20240101/00/atmos/gdas.t00z.atmanl.nc", ...]
+    """
+    def assert_datetime_format(date_str, fmt="%Y%m%d%H", label="date"):
```
CoryMartin-NOAA: If using the ISO format in the YAML:
datetime.fromisoformat(date_string) should do the trick and simplify things?

src/eva/tests/config/testMonSummary.yaml
EdwardSafford-NOAA: That would make it consistent with eva's time-series mechanism so it's probably the way to go.

src/eva/tests/config/testMonSummary.yaml
```diff
@@ -14,8 +14,8 @@ datasets:
       - ${data_input_path}/time.hirs4_metop-a.ctl
     filenames_template:
         template: ${data_input_path}/time.hirs4_metop-a.%Y%m%d%H.ieee_d
-        start: "2015051418"
-        end: "2015051518"
+        start: "2015-05-14T18:00:00"
+        end: "2015-05-15T18:00:00"
```
CoryMartin-NOAA:
```suggestion
        start: "2015-05-14T18:00:00Z"
        end: "2015-05-15T18:00:00Z"
```

src/eva/utilities/utils.py
```diff
@@ -548,18 +533,25 @@ def assert_datetime_format(date_str, fmt="%Y%m%d%H", label="date"):
...
             raise ValueError(message)
 
+    # Assert start and end time are in ISO format
+    try:
+        start = datetime.fromisoformat(config["start"])
+        end = datetime.fromisoformat(config["end"])
+    except ValueError as e:
+        message = (
+            f"Datetime format: {e}. "
+            "Use format like 'YYYY-MM-DDTHH:MM:SS'"
```
CoryMartin-NOAA:
```suggestion
            "Use format like 'YYYY-MM-DDTHH:MM:SSZ'"
```

kevindougherty-noaa: From what I have gathered, `datetime.fromisoformat()` does not work with Z at the end. It throws a value error.

CoryMartin-NOAA: Looks like it is a python 3.11 addition. Two suggestions: either change your example doc string (which includes Z) or modify the code so that if it ends in Z, strip it off the end

CoryMartin-NOAA (approved):

PR #248, opened 2025-08-28
Bugfix/regex arithmetic

Dooruk:
The regex change `abs` was replacing `absoluteDynamicTopography` from the ocean obs. It took me a bit to identify this bug..

### DIFF
--- a/requirements_sles15.txt
+++ b/requirements_sles15.txt
@@ -1,20 +1,20 @@
 setuptools>=59.4.0
 pyyaml>=6.0
-pycodestyle>=2.8.0
+pycodestyle>=2.11.0
 netCDF4
 matplotlib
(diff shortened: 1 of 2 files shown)


Dooruk:
Sorry to bug you for this bug @CoryMartin-NOAA but I need write approval.

Dooruk:
Also @mranst apparently this was my problem with EVA, not the package versions.

https://github.com/GEOS-ESM/swell/pull/549

karpob:
This looks reasonable to me, I feel like there must be a better way to do what arithmetic does in general. Using `regex` for this seems like there may be unintended consequences based on what the user may input.

karpob (approved):

CoryMartin-NOAA (approved):

PR #250, opened 2025-10-01
Update all EMCPy hashes

kevindougherty-noaa:
## Description

In my last PR, I only updated one requirement file for the new EMCPy hash. I believe I need to update them all.

Alos had to make adjustments to how eva handled the EMCPy map gridded plotting. All tests are passing as expected.

### DIFF
--- a/requirements-github.txt
+++ b/requirements-github.txt
@@ -12,4 +12,4 @@ bokeh<3.6.0,>=3.5.0
 geopandas>=0.13.2
 geoviews>=1.10.0
 nbsite>=0.8.1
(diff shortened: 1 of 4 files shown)


kevindougherty-noaa:
@CoryMartin-NOAA @EdwardSafford-NOAA The changes to the map gridded plot are due to some tests inputting 1D lat/lon data and then looking for 3D data (i.e levels, lat, lon). So I had to add a bandaid that essentially rewrites data to 2D so the changes in EMCPy don't break. It has to do with how the plotting handles the cell centers and edges.

kevindougherty-noaa:
@CoryMartin-NOAA Do you mind taking a look at this?

CoryMartin-NOAA (approved):

